PR #3, opened 2015-02-19
Initial pass at /v2/account

lye:
The account endpoint is intended to provide access to all information about an account. It serves several sets of data; which sets are returned depend on which scopes the passed authentication token has access to. This endpoint unconditionally requires the "account" scope, so the basic account information is always returned. The returned sets are all merged into a single result object, so the end result is that with more scopes you'll have more fields in the response. Some example return sets --

 * Basic account information -- account "id", "name" and "world" (scope "account").
 * Bank and material storage inventories -- "bank" and "materials" (scope "inventories").
 * Wardrobe/dye unlocks (scope "fashion?" maybe)
 * Achievements/collections.
 * etc.

For this addition, since it's the first authenticated endpoint, we're only adding the most basic of endpoints for now -- the basic account information -- and the bank/materials fields are intended for demonstration purposes.


soulweaver91:
The simple example given on the bank data is a bit ambiguous. What if the player has an item on the, say, 2nd and 4th slots of their bank? The impression I get is that it would be like
```
// ...
  "bank" : [
    {
      "id" : 12178,
      "slot" : 2,
      "count" : 250
    },
    {
      "id" : 19677,
      "slot" : 4,
      "count" : 2
    }
  ],
// ...
```
The slot number is a part of each item though, which seems a bit counterintuitive since arrays already have an implicit equivalent (the item's position in it). The example also leads to believe (correct if I'm wrong) that indexing for the slots would start at 1 when most data structures are traditionally indexed from 0. I would probably find the following alternative structure a bit better:
```
// ...
  "bank" : [
    {
      "id" : null,
      "count" : 0
    },
    {
      "id" : 12178,
      "count" : 250
    },
    {
      "id" : null,
      "count" : 0
    },
    {
      "id" : 19677,
      "count" : 2
    },
    // either stop here or repeat with nulls until the end of available bank slots
  ],
// ...
```

Of course, there might be technical, design or even resource constraints here at play. I could see both versions working, some better in certain use cases and some not. Just throwing an idea :p

lye:
Your first example is correct -- if an item is contained in multiple slots, there will be two entries in the `bank` array containing that `id`.

In the current implementation, `bank` and `materials` are sparse arrays (e.g., "id" is never null and "count" is never 0). So your second example is currently encoded (using 1-indexed slots) as --

```json
{
    "bank" : [
        {
            "id" : 12178,
            "count" : 250,
            "slot" : 1
        },
        {
            "id" : 19677,
            "count" : 2,
            "slot" : 4
        }
    ]
}
```

I'm not really attached to either encoding -- it was just how we store it internally. Using placeholders to make the `slot` field implicit *seems* like it would make consumption easier in any case where you're displaying the contents in a UI and is trivial to strip out if you don't care about their positions (though the same case can be made in both ways).

Arguably, if you don't care about the slots you probably just want the total count of each item, which means you're transforming the data anyway.

lhw:
Bank, Materials, Dyes, Skins should definitely be different endpoints.
e.g. /v2/acccount/bank /v2/account/materials etc. Not on the same endpoint. That would be a lot of overhead if they were on the same endpoint and you just want to know which world the user is from.

lye:
Okay, wow, I had a brainfart while writing the original version.

 * Slots are 0-indexed, as you'd expect.
 * Objects in `materials` don't have `slot`, since they can't be moved around.

lye:
> Bank, Materials, Dyes, Skins should definitely be different endpoints.

Hrm, what if we provided `/v2/account` which dumped all of the data, and `/v2/account/{bank,materials,...}` which returned only the specific requested section?

Alternatively, what if `/v2/account` was bulk-expanded from string keys, e.g.

```js
// GET /v2/account
// Authorization: Bearer token-asdf

[ "info", "bank", "materials" ]
```

```js
// GET /v2/account?ids=bank,materials
// Authorization: Bearer token-asdf

{
    "bank" : {...},
    "materials" : {...}
}
```

tivac:
Using the bulk-expansion stuff feels a little weird to me here, but has the nice benefit of allowing you to specify exactly the data you need & mashing it all into a single request.

I'm torn. :confused:

lhw:
That seems really weird to be honest. I don't really see an issue with having different endpoints for the data. They have widely differing schemas and usage scenarios. I don't think they should be bundled up.

soulweaver91:
@lye: I agree with that transforming the array from either case to another is rather easy and thus it doesn't matter *that* much. Might just be that I'm not used to sparse arrays myself. :) One small piece of data my suggestion (with no array truncation) would automatically convey would be the total number of bank slots the player currently has unlocked - though of course, that doesn't need to be tied to that array either and could be a separate field as well. I was a bit curious about the slot field in the material storage, too, but I see you already removed that there :p

As for bulk expansion, I don't have a strong opinion either for or against it.

darthmaim:
Let's think about the use cases for those endpoints. Either you want to check if a user has that item in the vault or you want to display it like it is displayed ingame.

It would make sense to __include items with count 0__ in the material endpoint, so we can also display the items the account doesn't own without keeping lists of them that every developer had to keep up to date. I would even go as far as to __separate them into the categories__ we have ingame.

Since the material endpoint doesn't have slots, we could also just return a object with `item_id` as key and `count` as value.

So the material endpoint would look something like this:
```js
// GET /v2/account/materials
// Authorization: Bearer token-asdf

{
  "common_crafting_materials": {
    "1234": 0,
    "2345": 250,
    "1337": 42,
    //...
  },
  "fine_crafting_materials": { /*...*/ },
  "rare_crafting_materials": { /*...*/ },
  "ascended_materials": { /*...*/ },
  "gemstones_and_jewels": { /*...*/ },
  "cooking_materials": { /*...*/ },
  "festive_materials": { /*...*/ },
}
```
This can also work with bulk expansion to only get the sections we need (`GET /v2/account/materials/cooking_materials,festive_materials`).

----
For the bank endpoint i would also include the total number of slots available to the user.
```js
// GET /v2/account/bank
// Authorization: Bearer token-asdf

{
  "slots": 60
  "items": [
    { "id": 12178, "slot": 2, "count": 250 },
    { "id": 19677, "slot": 4, "count": 2 },
    //...
  ]
}
```

Ruhrpottpatriot:
@darthmaim I don't think bundling the materials in separate categories is the right way to go. I'D prefer the same style used in the items endpoint (i.e. a field telling me which category the material is)
I think it should look more like this:
```javascript
"materials" : [
   {
      "id" : 19699,
      "type" : "common",
      "count" : 250
   },
   {
      "id" : 19699,
      "type" : "common",
      "count" : 0
   }] 
```
Whereas each material has an entry in the list. Meaning if a user does not have a material in store it's count is set to zero

As for the "normal" bank. I'D like to see the second example of @soulweaver91. Although I could live with sparse arrays, too.

darthmaim:
I like `"type": "common"` for the materials, is closer to the bank and wouldn't break if new fields get added later.

dr-ishmael:
For materials - what about, instead of encoding the material category within every item in the account response, there was a separate endpoint /v2/materials that returned the current structure of the materials categories?  That response could be structured like darthmaim's first suggestion, i.e. category -> {id, slot}, and it would give you the complete structure instead of just a partial structure of the specific items that a certain account actually has popualted.

Ruhrpottpatriot:
I don't think adding a separate materials endpoint is a good idea. After all materials can be thought as a subset of a bank, only with less customisation options (items have a set position, etc.) Therefore it should be query able if you query the bank.
So I'd like to see this:
 * "v2/[token]/account" gives a complete overview over the account of [token]
* "v2/[token]/account/bank" gives a complete view of the standard bank
* "v2/[token]/account/materials" gives a view of the materials bank

dr-ishmael:
@Ruhrpottpatriot I'm talking about a generalized endpoint similar to `v2/files`, not an account-specific one.

* `v2/materials` gives the complete overview of all the material slots that currently exist in the materials bank. This gives the `id`, `category`, and `slot` number within the category for each material.
* `v2/account/` or `v2/account/materials` gives the limited-scope view of only the materials that the authenticated account has 1+ stored in their bank.  This only gives the `id` and `count` for each item the account has stored.

You would store the result from `v2/materials` and use it to build a template display of the materials bank - you save this template until a new build, when you should query it again to see if anything changed.  Then you get the account-specific data from `v2/account/materials` and use it to populate the display template.

Ruhrpottpatriot:
@dr-ishmael I don't think this is the right way to go. `v2/files` by itself has no relation to anything in the game. It is just a list with images.
On the other hand a `materials` endpoint has a clear relation with a user. No material exists by itself in the game, it is always bound to a user. Therefore I think those two should be linked.

But it seems I have not made myself clear enough. The `v2/[token]/account/materials` should always return an array of all materials, where the position in the array is the position in the bank itself.
Like this:
```javascript
    "materials" : [
    {
        "id" : 19699,
        "type" : "common",
        "count" : 250
    },
    {
        "id" : 19699,
        "type" : "
        "count" : 0
    }] 
```


A call to `v2/[token]/account` would then look like this:

```javascript
// GET /v2/account
// Authorization: Bearer oauth2-token

// With the "account" scope:
{
	"id"    : "account-guid",
	"name"  : "Lawton.1234",
	"world" : 1007
}

// With the "account" and "inventories" scopes:
{
	"id"        : "account-guid",
	"name"      : "Lawton.1234",
	"world"     : 1007
	"bank"      : [
		{
			"id"       : 19699,
			"count"    : 250
		},
		{
			"id"       : 30692,
			"count"    : 1,
			"upgrades" : [ 24607 ]
		}
	],
    "materials" : [
    {
        "id" : 19699,
        "type" : "common",
        "count" : 250
    },
    {
        "id" : 19699,
        "type" : "
        "count" : 0
    }] 
}
```

Ruhrpottpatriot:
Seems like @lye has updated the schema and I like it.

lye:
I've updated the PR with my personal preference from the suggestions so far.

For the bank, I think it makes more sense to include empty slots. This lets an application take the input array verbatim as a viewmodel without having to add in the empty slots by hand. In the case where you only care about the item counts, you'd still have to process the data to collect duplicate stacks. Having empty slots also lets us implicitly provide the total number of slots available.

Might be better to represent the bank items as an array of arrays -- so that it's clear where the delineation between slots lies -- but I'm not sure if that's helpful.

For the material storage, I'm kind of torn between having an object with keys for each storage category and just putting the storage category in the item objects. My logic for the latter is that it makes it easier to write forward-compatible application code in the unlikely case additional storage types are added.

It's not clear to me what the benefit of having a separate `/v2/materials` API is -- I guess this would allow us to provide additional metadata about the material categories (e.g., the localized heading name) -- but it makes determining the slot of each item somewhat less accessible than just providing an array with explicit empty slots, IMO.

Ruhrpottpatriot:
@lye Yeah I guess I am *lol* (or maybe it was chance after shelving some code for GW2.NET and checking back here)
But I really like your take on the endpoint. And I second putting the material category in the item object. This makes it easier should the case arise when there is a new category and it also is more consistent with existing API endpoints (items I'm looking at you over there).
I really don't see the benefit in adding a separate `v2/materials` endpoint. For Localisation we could do with a `?lang=xx` URL parameter like every other api so far.
So for consistency sake I give :+1: to your suggestion.

KhaymanAlisti:
Can the bank schema include info on to which character an item is soulbound?

codemasher:
@KhaymanAlisti This isn't even possible ingame and i wonder why. We had that already back in GW1: "customized for [character name]".

KhaymanAlisti:
@codemasher Aye; that discussion of why is probably better left for the forums. Just wondering if it is technically doable, and if it pushes for the v2/character interface sooner ;)

lye:
> Can the bank schema include info re: to which character an item is soulbound?

It can, but that feature probably won't make it for the initial release, so I'm going to leave it out for now.

lye:
Okay, after looking into it a bit, the `"type"` field isn't going to work as-is. First, we don't have a local symbol (aside from a meaningless number) and I really want to keep the implementation data-driven. Also there are issues with i18n (e.g., the symbol/number doesn't express the localized name of the type) I think the first suggestion is way better:

> For materials - what about, instead of encoding the material category within every item in the account response, there was a separate endpoint /v2/materials that returned the current structure of the materials categories?

This has the fortunate side-effect of making the material storage lists available without authentication (what was I even thinking).

So I'm going to change the response of `/v2/account/materials` to have objects that have a `category` field which can be cross-referenced against `/v2/materials` (or something along those lines -- I'm not really happy with that as the endpoint name). I'll update the examples tomorrow. Sorry for being flippy-floppy on this :<

lye:
Splitting off the `/v2/account/bank` and `/v2/account/materials` into separate PRs. Gonna merge this one with just the `/v2/account` since there haven't been any terrible flaws found in the bare-minimal functionality.

PR #4, opened 2015-02-19
Add basic authenticated /v2/character API for POST requests

Samuirai:
I believe the GW2 API is great, but too passive. This change would allow players to actively interact with their characters.

The suggested API has it's endpoint in `/v2/characters/<auth-token>/<character id>` and is accessible via POST request. It's obvious that a valid authentication token from the auth API is needed. The usage is an assumption so it is likely to change once the auth API is fully revealed.

example usage:

    POST /v2/characters/1234abcdefgh-auth-token/1
    { "action": "add_gold", "amount": 100000 }

example response:

    {
      "status": "success"
    }

Additional example requests can be seen in `v2/characters/character.js`

### DIFF
--- a/v2/characters/character.js
+++ b/v2/characters/character.js
@@ -1,0 +1,18 @@
+// POST /v2/characters/1234abcdefgh-auth-token/1
+// body:
+// { "action": "add_level", "amount": 1 }
+
+// POST /v2/characters/1234abcdefgh-auth-token/1
+// body:
+// { "action": "add_gold", "amount": 100000 }
+
+// POST /v2/characters/1234abcdefgh-auth-token/1
+// body:
+// { "action": "add_karma", "amount": 10000000 }
+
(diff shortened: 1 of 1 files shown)


lye:
Due to complex technical issues involving concurrent writes in a distributed system, we only have a read-only copy of character data. As such, we can't implement any endpoints like this.

poke:
What exactly is this action supposed to do? Also, this is very REST-unlike. Suppose there existed some “add gold” thing for characters (though I wonder where that gold comes from?), it should be something like `POST /v2/character/<character-identifier>/gold`.

But as stated before, authenticated requests are just coming, so there is a very good chance that we will eventually see API endpoints that allow us to perform actions instead of just querying data. It’s just that changing content requires a lot more testing than just getting data, so it makes sense to gradually expand it as the authenticated API proves itself safe.

Samuirai:
It was meant as a joke. But thank you for those serious responses :D

PR #5, opened 2015-02-19
Initial pass at /v2/skills

lye:
To preface this, skills are a bit complicated to represent fully. This is a first pass which completely omits what the skills do and how traits affect skills; we'll tackle that complexity in another PR. The intent for this PR is to expose the fields that aren't affected by traits so, if nothing else, we can get the skill names, descriptions and icons sorted.

`skills.js` contains a list of tentative examples, but I might be missing some important category that would have a weird representation -- if so, let me know and I'll paste in what the current code is outputting.

In general, a skill has the following fields:

 * `name` contains the localized name of the skill.
 * `description` contains the skill description.
 * `type` is one of `Weapon`, `Elite`, `Heal`, `Profession`, `Utility`, `Toolbelt`, `Transform` or `Bundle`.
 * `weapon_type` is one of the weapon types defined by the [/v2/items endpoint](http://wiki.guildwars2.com/wiki/API:2/items#Weapon).
 * `professions` is the list of professions that can equip the skill.
 * `slot` is one of `Elite`, `Heal`, `Weapon_{1-5}`, `Pet`, `Profession_{1-4}`, `Utility`, `Downed_{1-4}`, `Transform_{1-5}`.
 * `icon` is a URL to render.guildwars2.com.
 * `id` is the skill id.
 * `categories` is the list of categories the skill belongs to (e.g., `Shout`, `Mark`).

There may be additional fields depending on the skill:

 * `flip_skill`, if present, contains a reference to what skill the skill toggles to, e.g. [Temporal Curtain](http://wiki.guildwars2.com/wiki/Temporal_Curtain) toggles to [Into the Void](http://wiki.guildwars2.com/wiki/Into_the_Void), which is a distinct skill.
 * `next_chain`, if present, is the next skill in the skill chain.
 * `prev_chain`, if present, is the previous skill in the skill chain.
 * `cost` is the profession cost of the skill (e.g., Warrior Adrenaline).
 * `initiative`, if present, is how much Thief initiative the skill requires (not sure why this isn't cost?)
 * `attunement`, if present, is the elementalist attunement associated with the skill.
 * `transform_skills`, if present, contains a list of skill ids that will completely replace your entire skill bar.
 * `bundle_skills`, if present, contains a list of skill ids that will replace just your weapon bar.
 * `toolbelt_skill` indicates the toolbelt skill associated with Engineer utilities.
 * `dual_wield` is what Thief offhand is required to use the skill.

There might be some things I've missed; again if you'd like to see the current output for a given skill let me know and I'll paste it in.

### DIFF
--- a/v2/skills/skills.js
+++ b/v2/skills/skills.js
@@ -1,0 +1,238 @@
+// GET /v2/skills
+// No parameters returns a list of IDs that can be bulk-expanded.
+
+[
+	1175,
+	5486,
+	5487,
+	...
+]
+
+// GET /v2/skills/5497
+// GET /v2/skills?ids=5497
+// Bulk-expansion works like other endpoints.
+
+{
+	name: "Flamewall",
+	description: "Create a wall of flame at the target area that burns foes.",
+	type: "Weapon",
+	weapon_type: "Focus",
+	attunement: "Fire",
+	professions: [
+		"Elementalist"
+	],
+	slot: "Weapon_4",
+	icon: "...",
+	id: 5497
+}
+
+
+// GET /v2/skills?page=0&page_size=1
+// Pagination included.
+
+[
+	{
+		name: "Bandage",
+		description: "Call your allies for help. Restores health while you channel it.",
+		type: "Weapon",
+		weapon_type: "None",
+		professions: [
+			"Guardian",
+			"Warrior",
+			"Engineer",
+			"Ranger",
+			"Thief",
+			"Elementalist",
+			"Mesmer",
+			"Necromancer"
+		],
+		slot: "Downed_4",
+		icon: "...",
+		id: 1175
+	}
+]
+
+// GET /v2/skills?ids=5534
+// Some skills have weird fields -- transform_skills refers to a skill bar
+// that replaces your weapon skills. The list contains skill ids.
+
+{
+	name: "Tornado",
+	description: "Transform. Gain stability and shape-shift into a tornado that damages and launches foes.",
+	type: "Elite",
+	weapon_type: "None",
+	professions: [
+		"Elementalist"
+	],
+	slot: "Elite",
+	categories: [
+		"Transform"
+	],
+	transform_skills: [
+		5752,
+		5753,
+		5754,
+		10586,
+		10586
+	],
+	icon: "...",
+	id: 5534
+}
+
+// GET /v2/skills?ids=5516,5517
+// bundle_skills is basically the same thing; the distinction (I think?)
+// is that you can weapon-swap away a bundle and still have access to your
+// utilities.
+
+[
+	{
+		name: "Conjure Fiery Greatsword",
+		description: "Conjure. Manifest a fiery greatsword in your hands and at the target location. When it lands, it damages and burns foes. Wielders of this weapon gain increased power and condition damage.",
+		type: "Elite",
+		weapon_type: "None",
+		professions: [
+			"Elementalist"
+		],
+		slot: "Elite",
+		categories: [
+			"Conjure"
+		],
+		bundle_skills: [
+			5532,
+			5533,
+			5697,
+			5517,
+			5531
+		],
+		icon: "..",
+		id: 5516
+	},
+	{
+		name: "Fiery Rush",
+		description: "Charge at your foe, leaving a line of fire behind you.",
+		type: "Bundle",
+		weapon_type: "None",
+		professions: [
+			"Thief",
+			"Mesmer",
+			"Warrior",
+			"Elementalist",
+			"Necromancer",
+			"Guardian",
+			"Ranger",
+			"Engineer"
+		],
+		slot: "Weapon_4",
+		icon: "...",
+		id: 5517
+	}
+]
+
+// GET /v2/skills?ids=5812,6111
+// Engineer utilities have a toolbelt_skill to indicate the associated 
+// skill. The flip_skill seems a bit weird, but indicates that "Bomb Kit"
+// turns into "Stow Bomb Kit" when in use.
+
+[
+	{
+		name: "Bomb Kit",
+		description: "Device Kit. Equip a kit that replaces your weapon with bomb skills.",
+		type: "Utility",
+		weapon_type: "None",
+		professions: [
+			"Engineer"
+		],
+		slot: "Utility",
+		toolbelt_skill: 5813,
+		flip_skill: 6111,
+		categories: [
+			"Kit"
+		],
+		bundle_skills: [
+			5842,
+			5823,
+			5822,
+			5824,
+			5939
+		],
+		icon: "...",
+		id: 5812
+	},
+	{
+		name: "Stow Bomb Kit",
+		description: "Stow your bomb kit.",
+		type: "Utility",
+		weapon_type: "None",
+		professions: [
+			"Engineer"
+		],
+		slot: "Utility",
+		toolbelt_skill: 5813,
+		icon: "...",
+		id: 6111
+	}
+]
+
+// GET /v2/skills/14375
+// Warrior burst skills have a "cost" to indicate the adrenaline cost.
+// Also note the slot.
+
+{
+	name: "Arcing Slice",
+	description: "Burst. Deliver a circular attack to foes around you, and gain fury. Deal more damage to foes with low health.",
+	type: "Profession",
+	weapon_type: "None",
+	professions: [
+		"Warrior"
+	],
+	slot: "Profession_1",
+	cost: 30,
+	flip_skill: 14545,
+	categories: [
+		"Burst"
+	],
+	icon: "...",
+	id: 14375
+}
+
+// GET /v2/skills/13005
+// Thief stealth skills have the "StealthAttack" category.
+
+{
+	name: "Backstab",
+	description: "Stealth Attack. Attack your foe from the shadows, striking for double damage if you hit from behind.",
+	type: "Weapon",
+	weapon_type: "Dagger",
+	professions: [
+		"Thief"
+	],
+	slot: "Weapon_1",
+	categories: [
+		"StealthAttack"
+	],
+	icon: "...",
+	id: 13005
+}
+
+// GET /v2/skills/13006
+// And normal thief weapon skills have their initiative cost listed in
+// the "initiative" field. For their dual-wield skills, the offhand weapon
+// type is listed in the dual_wield field.
+
+{
+	name: "Death Blossom",
+	description: "Dual Wield. Do an evasive attack over your target, striking nearby foes three times. Each hit causes bleeding.",
+	type: "Weapon",
+	weapon_type: "Dagger",
+	professions: [
+		"Thief"
+	],
+	slot: "Weapon_3",
+	dual_wield: "Dagger",
+	categories: [
+		"DualWield"
+	],
+	initiative: 4,
+	icon: "...",
+	id: 13006
+}


dr-ishmael:
`type` and `slot` seem almost redundant.  What about the skill types that determine trait interactions like `Signet, Shout, Trap, Symbol, Gadget, ...`?

lye:
Yeah, there's a pretty big terminology mismatch between what things are known as internally and how they're shown to players. It's kind of a hassle at times.

The existing `type` field is a bit ... duplicated with `slot` though. What if `type` was renamed to `source` and the old values were mapped as such:

 * `Elite`, `Heal`, `Utility` -> `Slottable`

And then `categories` was renamed to `types`? e.g.

```js
{
    name: "Mark of Horror",
    description: "Mark. Minion. Inscribe a mark that summons five jagged horrors when a foe triggers it.",
    source: "Weapon",
    weapon_type: "Staff",
    professions: [],
    slot: "Weapon_4",
    flip_skill: 19114,
    types: [
        "Mark",
        "Minion"
    ],
    icon: "...",
    id: 10636
}
```

Also, hmm, there needs to be a good way to indicate that a skill originates from a transform/bundle, because `professions:[]` for transform skills is kind of strange.

dr-ishmael:
`source` is actually how I've always thought of that piece of data, so those renames would work for me.

I've had time to think of a lot more questions now. :)

1. Could you add a set of chain skills to your examples?  I think those are the only "additional fields" you didn't cover.
1. Your `Bundle` examples are for an elementalist conjure and an engineer kit.  Will you include  skills for environmental bundles, e.g. [Wooden Plank](http://wiki.guildwars2.com/wiki/Wooden_Plank)?  Or the thief's [stolen skills](http://wiki.guildwars2.com/wiki/Stolen_skill#Stolen_skills), which (I'm pretty sure) are just invisible bundles?
1. Continuing from there, is there any way to tie those back to the bundle that grants them since it's not from another skill?
1. In a similar vein for `Transform` - will skills from activity- and story-related transforms ([Crab Toss](http://wiki.guildwars2.com/wiki/Crab_Toss), [Caithe transform](http://wiki.guildwars2.com/wiki/Caithe_(transformation))) be included?
1. I noticed that for the non-weapon skills in your examples, they all have `weapon_type: "None"`.  Since that attribute obviously only applies to weapon skills, is it possible to suppress it for non-weapon skills?
1. On the warrior burst skills, I assume there are 3 (or 4) different versions of each skill with different `cost` values that correspond to the stages of the adrenaline bar?
1. Going over to traits now (I know you said they'll be covered in another PR, but I just want to ask this), we had a user on the wiki document skill IDs a while back, and it appears that when a trait affects a skill, it "flips" it to a different version (e.g. in his list of [Engineer skills](http://wiki.guildwars2.com/wiki/User:Malgalad/0x07_codes/Engineer), the unmodified Hip Shot is 5814, but switches to 6003 with the Rifled Barrels trait).  If this is correct, it doesn't seem like there would need to be much done in the skills endpoint to support traits; the trait's info would list which skills it causes to switch to a different version.

lye:
> Could you add a set of chain skills to your examples? I think those are the only "additional fields" you didn't cover.

```js
{
    name: "Smack",
    description: "Chain. Smack your foe. Repair turrets.",
    type: "Bundle",
    weapon_type: "None",
    professions: [
        "Engineer"
    ],
    slot: "Weapon_1",
    next_chain: 5993,
    flip_skill: 5993,
    categories: [ ],
    icon: "...",
    id: 5992
},
{
    name: "Whack",
    description: "Chain. Whack your foe. Repairs turrets.",
    type: "Bundle",
    weapon_type: "None",
    professions: [
        "Engineer"
    ],
    slot: "Weapon_1",
    next_chain: 5994,
    prev_chain: 5992,
    flip_skill: 5994,
    categories: [ ],
    icon: "...",
    id: 5993
},
{
    name: "Thwack",
    description: "Thwack your foe. Repairs turrets.",
    type: "Bundle",
    weapon_type: "None",
    professions: [
        "Engineer"
    ],
    slot: "Weapon_1",
    prev_chain: 5993,
    icon: "...",
    id: 5994
}
```

TBH, I'm not entirely happy with how the skill chains are encoded. An array of skill IDs might be better.

> Will you include skills for environmental bundles, e.g. Wooden Plank? Or the thief's stolen skills, which (I'm pretty sure) are just invisible bundles?
> In a similar vein for Transform - will skills from activity- and story-related transforms (Crab Toss, Caithe transform) be included?

There's no sane way to determine if a skill is player-usable, so right now I'm using some heuristics combined with manual whitelisting to figure out what skills should be returned. As such, I think both environmental bundles and transforms are out-of-scope (as it's much harder to verify whether or not they can be acquired). A Thief's stolen skills *should* be included, but are currently not (must be something filtering them out, will have to investigate).

> Continuing from there, is there any way to tie those back to the bundle that grants them since it's not from another skill?

There needs to be, at least for conjures/kits, because it's not really clear ATM. Ditto for transformations.

> On the warrior burst skills, I assume there are 3 (or 4) different versions of each skill with different cost values that correspond to the stages of the adrenaline bar?

Yeah, there's a separate skill for each stage. I wasn't sure how to encode that meaningfully, so I'm only showing the first stage. Might make sense to remove the `cost` entirely since it can be effected by traits, and I'm trying to avoid any such fields until we come up with a sane representation for the modifications made.

>  If this is correct, it doesn't seem like there would need to be much done in the skills endpoint to support traits; the trait's info would list which skills it causes to switch to a different version.

Only some traited skills use a separate skill -- others modify the same skill in-place. This can be seen in-game by a "flip" animation when you trait skills. In general, anything that modifies range/cooldown/cost will be a separate skill, which additional/modified effects will usually not be a separate skill.

dr-ishmael:
> There's no sane way to determine if a skill is player-usable, so right now I'm using some heuristics combined with manual whitelisting to figure out what skills should be returned. As such, I think both environmental bundles and transforms are out-of-scope (as it's much harder to verify whether or not they can be acquired). A Thief's stolen skills should be included, but are currently not (must be something filtering them out, will have to investigate).

I understand the player-centric view, since the primary usage for this endpoint will be to feed build calculators.  Personally, I'd love to have the data on every single skill/effect in the game so that the wiki can finally have accurate documentation on all the NPC-only abilities.

>> Continuing from there, is there any way to tie those back to the bundle that grants them since it's not from another skill?

> There needs to be, at least for conjures/kits, because it's not really clear ATM. Ditto for transformations.

When the parent skill is player-usable, like Lich Form, I don't know if it's necessary to have a "backlink" from the bundle/transform skill to the parent skill (although some people would find it useful, I'm sure) because you can read the `transform_skills` or `bundle_skills` from the parent.  Since environmental bundles and contextual transforms won't be included, this doesn't worry me anymore.

> Yeah, there's a separate skill for each stage. I wasn't sure how to encode that meaningfully, so I'm only showing the first stage. Might make sense to remove the cost entirely since it can be effected by traits, and I'm trying to avoid any such fields until we come up with a sane representation for the modifications made.

Hmm... if it's possible, I think translating `cost` into `adrenaline_stage: ( 0 | 1 | 2 | 3 )` would be a nice, meaningful way to present it.

> Only some traited skills use a separate skill -- others modify the same skill in-place. This can be seen in-game by a "flip" animation when you trait skills. In general, anything that modifies range/cooldown/cost will be a separate skill, which additional/modified effects will usually not be a separate skill.

Ooooooh... yeah, okay, that's definitely too complicated for a first-pass discussion. :)

Ruhrpottpatriot:
I don't like how the slot is encoded. For me it violates one important principle: "Atomicity". For example the `Mark of Horror` skill should look like this:
```
{
    name: "Mark of Horror",
    description: "Mark. Minion. Inscribe a mark that summons five jagged horrors when a foe triggers it.",
    type: "Weapon",
    weapon_type: "Staff",
    professions: [],
    slot: 4,
    flip_skill: 19114,
    categories: [
        "Mark",
        "Minion"
    ],
    icon: "...",
    id: 10636
}
```
Why? It is not important to know if the slot is `Elite`, `Heal`, `Weapon_{1-5}`, `Pet`, `Profession_{1-4}`, `Utility`, `Downed_{1-4}` or  `Transform_{1-5}`. The slot is the position where the skill in the skill bar is located i.e. 0-9 (or 1-10). The other thing is the type of the skill, which should be encoded in it's own field. Also I don't like the name of the field source. Yes it describes where the skill comes from, but does a skill really has the "source" `Elite`? Source really only makes sense if we are talking about a weapon skill, every other type is somewhat odd.
Further I liked the old naming of `types` much more. While it still is a good name it would conflict with the `type` field above. I think of it as categories. I can order the skill in *n* categories. Note, that this is manly semantics but it somehow feels more... natural to me at least.

Here is another example:
```
    name: "Thwack",
    description: "Thwack your foe. Repairs turrets.",
    type: "Bundle",
    professions: [
        "Engineer"
    ],
    slot: 1,
    prev_chain: 5993,
    icon: "...",
    id: 5994
```

Kuliya:
What about an output for what type of targeting the skill is: Direct, ground targeted, shout, path directed, self.
http://wiki.guildwars2.com/wiki/Targeting

Also, what about the type of field it is: Fire, water, etc.

Or what kind of interaction you can expect from using said skill (IE Finisher): leap, whirling, blast, projectile.
http://wiki.guildwars2.com/wiki/Combo

lye:
> It is not important to know if the slot is Elite, Heal, Weapon_{1-5}, Pet, Profession_{1-4}, Utility, Downed_{1-4} or Transform_{1-5}. The slot is the position where the skill in the skill bar is located i.e. 0-9 (or 1-10)

That actually makes a lot of sense -- I'll fiddle with this some bit and see what it looks like. There might be some edge-cases where it breaks down a bit but not really sure.

> What about an output for what type of targeting the skill is: Direct, ground targeted, shout, path directed, self.

I'm intentionally omitting any attribute that can be changed via traits; we can discuss that in a future pull request. Not sure that excuse works for fields, but it [might work for finishers](http://wiki.guildwars2.com/wiki/Stronger_Bowstrings).

Ruhrpottpatriot:
@lye The only place where I could think of a break down would be the F1 - Fxyz, but for those I'd put the position either as 10+ (given a start index of 0), or encode the position as string and encode the bar skills as A0-9 and the above bar skills (for a lack of better wording) as F0-3.

Or are there some super complicated skills that fall out of the Roster.

Oh wait... Monster skills. However I'd simply do the following for them. Since monster do not have an ability bar, just set the location to -1. With this we can label every skill which has no position in or above the ability bar.

lye:
> @lye The only place where I could think of a break down would be the F1 - Fxyz, but for those I'd put the position either as 10+ (given a start index of 0)

Ehh, if we still have the `type`/`source` field then `"Profession"` would be there, and `"slot"` would correspond to the 1-4 in F1-F4. I think it still holds.

> Oh wait... Monster skills. However I'd simply do the following for them. Since monster do not have an ability bar [...]

Well, they actually do have an ability bar of sorts. Also I'm not going to expose monster skills via `/v2/skills` -- they're not designed to be player-facing at all (and there's no sane way to determine which monster skills are even encounterable).

TimeBomb:
Nice, I'm quite excited for this endpoint! It's looking great so far.

Is this API intended to require authentication? i.e. It shows information related to that specific player's current skills? Given this API does not contain "account", and that it is intended for consumption by skill calculators, I'll be going off the assumption "no".

- "type" replacing "categories" is quite alright.

- "slot" doesn't make much sense to me. There is no information about any player here, only abilitis. Players choose which ability to put in which slot. How does a skill have a designated slot?

- Given the above point, the current iteration of slot should be removed, and "type" should be renamed to "slot". Among other things, many people often refer to it ingame as such.

- I agree with you that "initative" should be merged with cost. I imagine adrenaline cost is shown as 0/1/2/3? That'd be nice.

- I'm not a big fan of "source". It's too vague. As Ruhrpottpatriot also pointed out, `source: "Profession"` doesn't make much sense. Keeping it as is would leave it in a better state.

- On one end, moving abilities like chains into arrays would be nice. A logical structure. On the other end, it would make using this API less consistent and harder to work with. If I'm iterating over a list of skills, it's nicer if I can expect them all to be consistent objects, rather than some of them be objects and some arrays. I can still traverse chains easily enough, given I have the previous/next skill ID.

- Food for thought: Revenants will be able to change their entire profession skill bar, or something to that degree, no? We'll either need something akin to `bundle_skills`, or we'll want to make `bundle_skills` slightly more abstract in definition. Judging by the way the API is currently designed, a new property will come about.

darthmaim:
> "slot" doesn't make much sense to me. There is no information about any player here, only abilities. Players choose which ability to put in which slot. How does a skill have a designated slot?
Weapon skills for example have a fixed slot and can't be moved by the player.

TimeBomb:
Ah of course, that makes sense for weapon skills. I fixated on the confusion around specific profession skill slots. Do any profession skills have specific slots?

lye:
> Do any profession skills have specific slots?

Mesmer F1-F4, etc.

TimeBomb:
Sounds good.

Ruhrpottpatriot:
>I agree with you that "initiative" should be merged with cost. I imagine adrenaline cost is shown as 0/1/2/3? That'd be nice.

While sounding good on paper this is not doable in the api. As @lye said:  An adrenaline skill is not one but multiple skills. A cost array wouldn't be able to reflect said skills.

dr-ishmael:
>> I agree with you that "initiative" should be merged with cost. I imagine adrenaline cost is shown as 0/1/2/3? That'd be nice.

> While sounding good on paper this is not doable in the api. As @lye said: An adrenaline skill is not one but multiple skills. A cost array wouldn't be able to reflect said skills.

What he means is 0 or 1 or 2 or 3 are the possible values for adrenaline cost, not that all of them are shown together on a single skill.

lye:
> What he means is 0 or 1 or 2 or 3 are the possible values for adrenaline cost, not that all of them are shown together on a single skill.

Ehhh, I kind of wanted them to be all shown together on a single skill, but I doubt it'll be feasible in a meaningful way. Probably should just toss the adrenaline cost into the "cost" field and leave it to the application logic to know that there are three versions of each burst skill.

TimeBomb:
@dr-ishmael Exactly, thanks for the clarification.

> Ehhh, I kind of wanted them to be all shown together on a single skill, but I doubt it'll be feasible in a meaningful way. Probably should just toss the adrenaline cost into the "cost" field and leave it to the application logic to know that there are three versions of each burst skill.
Ah I see, yes, that could be quite useful. As @Ruhrpottpatriot seemed to imply, I wouldn't be a big fan of a property that could have multiple different types depending on the ability, e.g. cost array or cost integer. Developers should be able to understand that it is a warrior skill, thus it is indeed adrenaline, and adrenaline always has a range between 0 and 3.

Alternatively, if it's very important to list how the ability works differently depending on adrenaline, there could be multiple of the same abilities, one for each possible adrenaline level.

Valento92:
Surprising that Ishmael didn't mention it, but the wiki set up SMW in a way that can help clarify this endpoint's returned structure, please check here http://wiki.guildwars2.com/wiki/Guild_Wars_2_Wiki:Semantic_MediaWiki/Skills.

Some parameters can be merged, like initiative and adrenaline (as suggested above), and some are expected to be multi-valued (type, chain sequence). I've been pulling data from the wiki through its SMW API, and came up with an expected parameters table, not sure if it helps but here you go: http://wiki.guildwars2.com/wiki/User:Valento/gw2wwrapper#Properties

Valento92:
Some things to be noted (inspired by SMW API):
- There should be a context parameter specifying to which a certain skill applies to: Weapon (slots 1 – 5), Healing (slot 6), Utility (slots 7 – 9), Elite (slot 0), Mechanic (slots F1 – F4), Bundle (slots 1 – 5), Secondary — any skill that requires another skill to be activated before it is available (chains, transforms/kits, sequences/toggles, etc.), Aquatic — any skill that is only available underwater, Downed, Drowning.
- Slot is only really meaningful for 1-5 skill bar (i.e. weapon, and bundle skills), since healing is always 6, utility can be slotted wherever the player wants (7-9), and elite is always 0. Other cases simply don't fit a slot parameter.

Let's try "Hold the Line!":
```javascript
{
    name: "\"Hold the Line!\"",
    description: "Grant protection and regeneration to allies.",
    context: "Utility",
    profession: "Guardian",
    types: [
        "Shout"
    ],
    icon-signature: "...",
    id: 10636
}
```

What about Symbol of Swiftness:
```javascript
{
    name: "Symbol of Swiftness",
    description: "Sear a mystic symbol into the target area, damaging foes and granting swiftness to allies.",
    context: "Weapon",
    profession: "Guardian",
    weapon: "Staff",
    slot: 3,
    types: [
        "Symbol"
    ],
    ground-targeting: true,
    icon-signature: "...",
    id: 9143
}
```

StrategistXIV:
I think this is already looking good! One thing we might want to add on this is the boons or conditions it gives and for how long. I believe a developer / player itself can figure out if it is outgoing or incoming (like some necro skills), but the durations aren't always shown in the straight skill fact. (Unless you play to add those things to the description as well.)

I could try to make a list for each profession on what specific fields they would need? (If people think that could be useful)

Eearslya:
Will this be updated to include the `skill_facts` object from #67?

lye:
Yeah, the skill objects will be exactly the same as the skills embedded in traits.

There's some other madness w.r.t. whitelisting skills (because I only want to expose the small subset that players can actually use) that requires a manual thumbing through still; it's a bit easier with traits since traits are a bit more uniform.

Lootballs:
Is there a expected release date (or a -beta) we can have a bash with? With traits and the character blob this is the last part of a build editor.

lye:
Hmm, there's a couple of major outstanding issues with the current
implementation -- the serialization is the same code as for the trait
skills -- but the skill enumeration bits are a bit wonky. I'll take another
look at it on Monday and report back.

StrategistXIV:
Were you able to look at it Lye? I wonder how this is implemented. Are you pulling the data directly from the game and then transposing it into JSON. Or do you have to manually create all the skills? Otherwise the community could maybe help? Do you think the skill API has a chance to be there anytime soon?

lye:
> Are you pulling the data directly from the game and then transposing it into JSON.

Yes. The component that sources this data basically just reads the .dat file.

I need to hand-verify that we're not exposing any skills that shouldn't be exposed -- which either means going through every skill and manually whitelisting them or implementing and testing some complicated heuristics (which I'm not wont to do because I'm a terrible programmer).

I mostly want to avoid accidentally spoiling the skills for the unreleased specializations and I'm pretty sure the current implementation would spoil 'em at some point in the future.

StrategistXIV:
Thanks for your response!:

Personally I think that not including skills that are for future releases are the best, however if this complicates everything too much and the fact that people will datamine the skills anyway, might be a reason to just include future skills as wel. Did tou mainly consider this out of data protection or of that of application breaking with non-existant skills?

Is there a way we can help? Maybe some of us could write Java, C or C++ code that could parse the current skills? Please feel free to let us know if we can help!

lye:
> Did you mainly consider this out of data protection or of that of application breaking with non-existent skills?

It's just the rule that we should attempt to not expose things before they're announced. That the skills get datamined anyway is beyond my perview; the only important bit to me is that the API isn't the one leaking.

> Is there a way we can help? Maybe some of us could write Java, C or C++ code that could parse the current skills?

Not really; it's all internal code. We do have [job openings](http://arena.net/#careers) though :)

Lootballs:
With all specializations announced, is there an ETA on this?

lye:
After HoT ships :O 

There's some nastiness to how skills are enumerated on the backend and fixing it either requires manually whitelisting all skills which should be emitted or throwing caution to the wind.

Valento92:
Wow, that will require quite a lot of time to get done correctly. Is the data structure that nasty to reason?

lye:
It just wasn't designed for public consumption, really.

Valento92:
I see. Hope you can come up with a nice endpoint structure!

Lootballs:
Any ETA on this now?

lye:
Hmm, might as well turn it on next week. I've had the unit tests for the endpoint disabled for awhile so I want to turn those on and make sure there's nothing horribly wrong, but aside from that I don't forsee there being any reason to leave it off. Thanks for the reminder (I have the brain of a goldfish).

lye:
Bleh, I lied. There's some bitrot in one of the backend pieces that needs to be corrected -- will have to wait until the next game release before I can turn the endpoint on.

Eearslya:
Just a couple things I noticed when documenting this for the wiki.

`6084`, `15800`, and `29722` are all engineer skills with a category of `Elixer` rather than `Elixir`.

And I'm not sure if this is intentional, but all elementalist conjure weapon skills have a `weapon_type` of `Sword` and no professions or categories.

lye:
> 6084, 15800, and 29722 are all engineer skills with a category of Elixer rather than Elixir.

Oof, thanks for the catch. 5PM is a good time for a deploy right?

> all elementalist conjure weapon skills have a weapon_type of Sword and no professions.

I need to look into that one a bit harder; I feel like they should be bundle types but there might be some other weirdness going on.

StrategistXIV:
Looks great! I am very happy to see this is live! Thanks lye! One thing that might be nice to add is skill coefficient of the skill btw! Maybe add that to the facts array too? (Since it can be overridden) But maybe I missed something:P

PR #6, opened 2015-02-20
Initial pass at /v2/commerce/transactions

lye:
`/v2/commerce/transactions` is a proposed authenticated endpoint to export a user's Black Lion Trading Post transaction data (e.g., the stuff in the "My Transactions" tab in-game). Since it uses the same backend as the in-game UI, it suffers from the same data retention problems -- at some point old data is purged from our systems (maybe 3 months?). This endpoint would allow 3rd party services to export and retain a user's transaction data.

It includes four distinct endpoints which all have the same response format:

 * `/v2/commerce/transactions/current/buys`
 * `/v2/commerce/transactions/current/sells`
 * `/v2/commerce/transactions/history/buys`
 * `/v2/commerce/transactions/history/sells`

These all do what they say on the tin. It's important to note that these endpoints are not bulk-expandable (i.e., they don't accept an `id`/`ids` parameter), but they *are* paginated and provide the appropriate response headers (e.g., `Link`, `X-Result-Count`, `X-Result-Total`, `X-Page-Size`, `X-Page-Total`).

### DIFF
--- a/v2/commerce/transactions/transactions.js
+++ b/v2/commerce/transactions/transactions.js
@@ -17,8 +17,6 @@
 
 // GET /v2/commerce/transactions/current/buys
 // GET /v2/commerce/transactions/current/sells
-// GET /v2/commerce/transactions/history/buys
-// GET /v2/commerce/transactions/history/sells
 
 [
 	{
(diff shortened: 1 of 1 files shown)


dlonie:
Just to clarify, the "current" end points will return unfilled buy orders and pending sales, while "history" will give completed transactions?

tivac:
@dlonie yes.

Does it make sense to have `/v2/commerce/transactions` to implicitly be unfulfilled transactions, and then you use `/v2/commerce/transactions/history` to get fulfilled transactions? Looking at it now I dunno how useful the `/current` part of the url is.

dlonie:
@tivac I think as long is it's clearly documented, either is fine :-)

My personal preference would be to opt for something more specific like 

`/v2/commerce/transactions/buy_orders`
`/v2/commerce/transactions/sell_orders`
`/v2/commerce/transactions/purchases`
`/v2/commerce/transactions/sales`

although the current endpoints have the advantage of matching the in-game TP UI.

tivac:
@dlonie I think matching the TP UI is valuable here in terms of clarifying what data the API will return.

dlonie:
@tivac I agree :)

Another suggestion, I agree that the `id`/`ids` parameters wouldn't be super useful here, but I'd really like to have the option to retrieve all listings in a certain time range. My use case will likely be to periodically query these endpoints to update a local copy of the data, and just grabbing all transactions since the last update would be super convenient.

Defaulting to say, the last month/week/day of data may make it less expensive to run the default query, as well.

lye:
> Defaulting to say, the last month/week/day of data may make it less expensive to run the default query, as well.

They're equally expensive -- we'd have to do a binary search of all results coming off the backend. There's probably a clever way to cache a search intermediaries such that the next "new data since January please" request from that user is faster, but that's perhaps being a bit too clever for my taste.

PR #2, opened 2015-02-19
New format for /v2/floors, moved under /v2/continents

tivac:
The current format used by /v2/floors is difficult for a couple of reasons.

* Sectors, Tasks, Skill Challenges, and POIs are arrays. Finding a specific element is at best O(n).
* You can't request only a specific map/region/poi/etc, and the responses for an entire floor across a continent can be very large (200KB+)
* Every sub-field of a map has it's own id key (`task_id` vs `poi_id` vs `sector_id`)

This PR contains my proposed changes to the format to fix those problems.

* `/v2/floors` has been removed. You access floor information under `/v2/continents` because a floor requires a continent ID anyways.
* the `points_of_interest`, `tasks`, and `sectors` for a map are now objects with elements keyed by their id. Looking up any of these items can now be done in O(1) time.
* `poi_id`/`sector_id`/`task_id` have now been standardized to `id`.
* If a POI is on the `default_floor` for the map it doesn't repeat that information now.

There's another part of this proposal that should help with the issue of response size. You can now drill down even further into these objects to get a more concise view of the data you care about. All of these new endpoints support the standard pagination params and headers to ease iterating through them and selecting just the data you care about.

```js
// GET /v2/continents/1/floors
// GET /v2/continents/1/floors?ids=all
// GET /v2/continents/1/floors?ids=1,2
// GET /v2/continents/1/floors?page=0
// GET /v2/continents/1/floors/1/regions
// GET /v2/continents/1/floors/1/regions?ids=all
// GET /v2/continents/1/floors/1/regions?ids=1,2
// GET /v2/continents/1/floors/1/regions?page=0
// GET /v2/continents/1/floors/1/regions/1/maps
// GET /v2/continents/1/floors/1/regions/1/maps?ids=all
// GET /v2/continents/1/floors/1/regions/1/maps?ids=1,2
// GET /v2/continents/1/floors/1/regions/1/maps?page=0
// GET /v2/continents/1/floors/1/regions/1/maps/26/pois
// GET /v2/continents/1/floors/1/regions/1/maps/26/pois?ids=all
// GET /v2/continents/1/floors/1/regions/1/maps/26/pois?ids=1,2
// GET /v2/continents/1/floors/1/regions/1/maps/26/pois?page=0
// GET /v2/continents/1/floors/1/regions/1/maps/26/tasks
// GET /v2/continents/1/floors/1/regions/1/maps/26/tasks?ids=all
[...]

PR #8, opened 2015-02-20
proposal for /v2/wvw/matches

codemasher:
So here's my proposal for the matches.json as per this wall of text: https://forum-en.guildwars2.com/forum/community/api/API-Suggestion-World-vs-World/3610535 (+following)

In summary:
* direct access to any matchup (no more loop through)
* UNIX timestamps
* > 2kb smaller in size compared to the v1 response

Ok, since i've messed up this pull request, i'm going to throw in the other wvw proposals:

match_details.json:
* complete overhaul
* detailed scores and objective stats for each map
* objectives in neutral state
```
    // GET /v2/wvw/match_details?ids=all 
    // GET /v2/wvw/match_details/2
    // GET /v2/wvw/match_details/2/6
    // GET /v2/wvw/match_details/2/6/red <-- overkill?
```
objectives.json:
* direct access to each objective
* names for each language
* map coordinates

world_names.json:
* direct access to each name
* names for each language

### DIFF
--- a/v2/wvw/match_details.js
+++ b/v2/wvw/match_details.js
@@ -1,0 +1,277 @@
+// GET /v2/wvw/match_details/2/8
+{
+  "match": {
+    // match_id could be safely dropped
+    "id": "2-8",
+    "region": 2,
+    "tier": 8
+  },
+
+  // time of last db update
+  "lastupdate": 1391927044,
+
+  // time of next score tick
+  "score_tick": 1391929000,
+
+  // colors should be ALWAYS green/blue/red
+  // according to the ingame rank order
+  // if so, this could be changed to an array:
+  // "maps": [2005, 2006, 2004],
+  // probably unnessecary if "id" is present for worlds/color
+  // however, it's convinient
+  "maps": {
+    "GreenHome": 2005,
+    "BlueHome": 2006,
+    "RedHome": 2004
+  },
+  "worlds": {
+    "red": {
+
+      // see "maps" comment
+      "id": 2004,
+
+      // last flip time for bloodlust possible?
+      "bloodlust": ["RedHome"],
+      "objectives": {
+        "camp": [
+
+          // each of these arrays could also carry the timestamp of the last flip
+          [39, ""],
+          [50, ""],
+          [51, "A7A5885B-B93D-4929-A196-ACDA578CF3DF"],
+          [52, ""],
+          [53, ""],
+          [5, ""],
+          [6, "A7A5885B-B93D-4929-A196-ACDA578CF3DF"]
+        ],
+        "castle": [],
+        "keep": [
+          [33, ""],
+          [37, "8E0F7987-559F-4C5D-A946-572868C4A649"],
+          [1, ""]
+        ],
+        "ruin": [],
+        "tower": [
+          [35, ""],
+          [36, ""],
+          [38, ""],
+          [40, ""],
+          [17, ""],
+          [20, ""]
+        ]
+      },
+      "scores": {
+        "BlueHome": {
+          "score": 5760,
+          "income": 0,
+          "camp": 0,
+          "tower": 0,
+          "keep": 0,
+          "castle": 0
+        },
+        "GreenHome": {
+          "score": 1252,
+          "income": 0,
+          "camp": 0,
+          "tower": 0,
+          "keep": 0,
+          "castle": 0
+        },
+        "RedHome": {
+          "score": 14432,
+          "income": 115,
+          "camp": 5,
+          "tower": 4,
+          "keep": 2,
+          "castle": 0
+        },
+        "Center": {
+          "score": 11410,
+          "income": 55,
+          "camp": 2,
+          "tower": 2,
+          "keep": 1,
+          "castle": 0
+        },
+        "Total": {
+          "score": 32854,
+          "income": 170,
+          "camp": 7,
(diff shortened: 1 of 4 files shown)


darthmaim:
I would prefer ISO 8601 for the dates instead of the unix timestamps, that is the de facto standard format for datetimes in JSON. It's readable, sortable and can easily be parsed by all languages/runtimes. ([and xkcd :smiley:](http://xkcd.com/1179/))

tivac:
Seconding @darthmaim's comment about timestamps.

I don't love the formatting for the objectives. I think it's gone too far in the direction of trying to be compact. We GZIP all our responses so bytes-over-the-wire isn't usually something I'm worried about. I'd rather build the API in such a way that it lends itself to making more specific requests for the data within it you need, like we did for [`/v2/floors`](https://github.com/arenanet/api-cdi/pull/2). In my ideal world there's no mystery data in the responses, and with this currently I had to go look at the `/v1` API to figure out what `[23, "<guid>"]` meant.

codemasher:
Ok, thats probably a cosmetic issue then.

```javascript
// ...
"objectives": {
  "camp": {
    "23": {
      "owner": "75FD83CF-0C45-4834-BC4C-097F93A487AF",
      "flipped": "whatever timestamp you want"
    }
  }
}
// ...
```

tivac:
@codemasher yeah, that's a lot clearer. Will try to carve out some time to look at the rest later today.

msmolev:
Does matches now need a new field to indicate type of the borderland? "New" kind that was demoed as a "replacement for current borderland" vs "traditional"? Or should this be determined by IDs of objectives present on the map (as in, if you have any of the "new" IDs then it's a new map vs traditional)

hdunderscore:
Can we get the world WvW rating, variance and volatility in the API ??

codemasher:
@hdunderscore i think this would fit best under the /v2/worlds endpoint. But yes, please!

tanis7x:
Is the `maps` array and/or the `worlds.red.id` field the ID of the map for `/v2/maps`? If so, I'd definitely like to to be `worlds.red.mapId` instead to make it a little more clear.Structurally, I think it also makes more sense within `worlds.[red|blue|green]`. 

I also want to extend the request from @tivac for more explicit labeling of information to `/matches`. I do like splitting out the region from the match ID though.

Something like this, perhaps:

    {
      "regions": {
        "NA": {
		  "startTime": 1391821200,
		  "endTime": 1392426000,
		  
		  "matches": {
            "1": {
			  "greenWorldId": 1019,
			  "blueWorldId": 1008,
			  "redWorldId": 1013
			},
            "2": {
			  "greenWorldId": 1016,
			  "blueWorldId": 1017,
			  "redWorldId": 1009
			},
		  }
        },
		
        "EU": {
		  "startTime": 1391796000,
		  "endTime": 1392400800,
		  
		  "matches": {
            "1": {
			  "greenWorldId": 2104,
			  "blueWorldId": 2010,
			  "redWorldId": 2203
			},
            "2": {
			  "greenWorldId": 2201,
			  "blueWorldId": 2202,
			  "redWorldId": 2301
			},
		  }
        }
      }
    }

With Heart of Thorns drawing nearer, it would be great if we could get the v2 WvW endpoints released so applications can be prepared for the new maps.

lye:
> worlds.red.id field the ID of the map for /v2/maps?

It should be, I think.

codemasher:
```worlds.red.id``` is the ID of the red world in this matchup. The map IDs are fix anyway for green/blue/red home. However, this could be dropped completely since we could assingn the data from ```/v2/matches```.

tanis7x:
The map IDs aren't really fixed- when Heart of Thorns is released, the map IDs will change for all borderlands.

Furthermore, [this blog post](https://www.guildwars2.com/en/news/essential-elements-of-the-new-borderlands/) states:

> Even though this will initially replace the existing Borderlands, we have plans to bring back the classic alpine Borderlands. More information on that will come in the future.

So in the future we will need to be able to swap out which map is being used for any matchup.

codemasher:
Even though the map IDs may change once with this update, the assignment of green/blue/red to the map ID will always be fix. So it just makes even more sense to have just the world ID in ```/v2/match_details``` instead of the map ID.

tanis7x:
I don't think the assignment of green/blue/red will be fixed.

As per the quote from that blog post, in the future we cannot guarantee that all matchups are using the desert borderlands. One week a world could be on the desert borderlands, and the next they could be on the alpine borderlands. We need that information somewhere. It would also be great if this could be in the API prior to the release HoT so that existing applications can support the switch as soon as it goes live.

It could easily go in `/v2/matches` as something like `borderlandsMapId`, or we could put it in the details. I don't have a particular preference which one, as long as it is somewhere.

bherbst:
@lye or @tivac any update or ETA for this?

As of right now the biggest challenges for me as a WvW app developer are that when Heart of Thorns is release I:

 * Will have no idea what data to expect from `/matche_details` or `/objectives`
 * Have no way of determining whether I should be using the alpine borderlands data or desert borderlands data (without relying on the release date for HoT in my code and assuming everything is desert from then on)
 * Do not know how to get desert borderlands map tiles from the tile API

lye:
I don't have an ETA, but I've got a version working locally. I'm gonna make a new PR for what I've got, will link it to here in a sec.

PR #12, opened 2015-02-27
Initial pass at /v2/account/bank

lye:
This is the `/v2/account/bank`, split off from #3.

### DIFF
--- a/v2/account/bank.js
+++ b/v2/account/bank.js
@@ -2,24 +2,70 @@
 // Authorization: Bearer oauth2-token
 // Requires "account" and "inventories" scopes.
 [
+	null,
 	{
-		"id"       : 19699,
-		"count"    : 250
+		id: 46774,
+		slot: 1,
+		count: 1,
+		upgrades: [
+			24675
+		]
 	},
 	{
-		"count"    : 0
+		id: 48087,
+		slot: 2,
+		count: 1,
+		skin: 89
+	},
+	null,
+	{
+		id: 46760,
+		slot: 4,
+		count: 1,
+		skin: 3706,
+		upgrades: [
+			24661
+		]
+	},
+	null,
+	null,
+	null,
+	null,
+	null,
+	{
+		id: 19996,
+		slot: 10,
+		count: 1
+		infusions: [
+			1234,
+			5678
+		]
 	},
 	{
-		"id"       : 19699,
-		"count"    : 12 
+		id: 19976,
+		slot: 11,
+		count: 4
 	},
-	{
-		"id"       : 30692,
-		"count"    : 1,
-		"upgrades" : [ 24607 ]
-	}
+	null,
+	null,
+	null,
+	null,
+	null,
+	null,
+	null,
+	null,
+	null,
+	null,
+	null,
+	null,
+	null,
+	null,
+	null,
+	null,
+	null,
+	null
 ]
 // Slot is indicated by array index; the number of slots is
-// implied by the length of the array. Empty slots are shown
-// with count=0 and id=undefined.
+// implied by the length of the array. Empty slots are indicated
+// by null values.
 
--- a/v2/account/bank.js
+++ b/v2/account/bank.js
@@ -1,0 +1,25 @@
+// GET /v2/account/bank
+// Authorization: Bearer oauth2-token
+// Requires "account" and "inventories" scopes.
+[
+	{
+		"id"       : 19699,
+		"count"    : 250
+	},
+	{
+		"count"    : 0
+	},
+	{
+		"id"       : 19699,
+		"count"    : 12 
+	},
+	{
+		"id"       : 30692,
+		"count"    : 1,
+		"upgrades" : [ 24607 ]
+	}
+]
+// Slot is indicated by array index; the number of slots is
+// implied by the length of the array. Empty slots are shown
+// with count=0 and id=undefined.
+


tivac:
I'd prefer if empty slots were just `null` instead of an object.

Ruhrpottpatriot:
@tivac hmmm NULL might be a good idea in Database context NULL means a value we currently do not know or does not exist after all.
However: The bank slot is existing it is just empty, so an object with a count of 0 is also reasonable. For deserializing it makes almost no difference if there is NULL, or an empty object. I can live with both, but prefer the current solution, since it is more intuitive.

StevenLiekens:
Guys! Guys... don't re-purpose array indexes for data storage. You can't guarantee that whatever json serialization engine is used will keep sort order of the source data.

lye:
> I'd prefer if empty slots were null instead of an object.

So I was thinking about this a bit more, and I changed my mind -- I like the explicit `{count:0}` because it's more consistent with the `/v2/account/materials` endpoint (where we *need* to represent it that way to include the item id, e.g., `{count:0, id:12138}`).

> don't re-purpose array indexes for data storage. You can't guarantee that whatever json serialization engine is used will keep the sort order of the source data.

Y-yes, we can. That's the definition of an array, it's an ordered container:

> An array is an ordered sequence of zero or more values ([RFC4627](http://www.ietf.org/rfc/rfc4627.txt))

StevenLiekens:
There's also the fact that not every language uses 0-based indexes. The first slot may have slot_id 0 or 1 depending on the language you're programming for.

darthmaim:
> There's also the fact that not every language uses 0-based indexes. The first slot may have slot_id 0 or 1 depending on the language you're programming for.

Both endpoints start at index 1 then, so it's not a problem. Also every developer working with a language that uses 1 based indexes knows that and can account for that.

Archomeda:
>So I was thinking about this a bit more, and I changed my mind -- I like the explicit `{count:0}` because it's more consistent with the `/v2/account/materials` endpoint (where we need to represent it that way to include the item id, e.g., `{count:0, id:12138}`).

@Iye Although it seems to be more consistent, a bank slot doesn't have a fixed item id, while material slots do, so these endpoints are not really the same. I think that having `{count:0}` is a bit useless. Having `null` works for me too. Although, on the other hand, if `count` is a guaranteed entry, developers can check against `count` if an item exists in that slot or not (`slot[i].count > 0`), instead of having to check if an item exists (`slot[i] != null`). Depending on the JSON library used, it might not even make that much of a difference.

I'm leaning towards the null-check myself, because basically, there is nothing, which is `null`. If there's a `{count:0}`, I would think that there is something, but just 0 items of it (which is actually perfect for /v2/materials etc).

>There's also the fact that not every language uses 0-based indexes. The first slot may have slot_id 0 or 1 depending on the language you're programming for.

@StevenLiekens To be honest, that's something a developer that uses this API should take into account himself. Since slot_id is not explicitly defined in this API (as far as I can see), the first slot is just the first slot, nothing else. The developer can even assign a starting slot_id of one thousand. It's up to him, it doesn't affect others.

lye:
@Archomeda, for me the compelling reason is that, in an application which displays a UI similar to the in-game UI, I'd want a view partial for representing a stack of an item (and it'd be the same partial between both the bank and material storage (and inventories, for that matter)). It's completely feasible to use `null` to indicate the same thing as `{count:0}`, but the calling code gets an extra 7 characters (!!):

```js
if (item && item.count) {..}
// vs
if (item.count) {..}
```

I'm really not that concerned about this detail though, I'll have a literal coin toss with @tivac to decide what empty untyped slots get serialized as. :laughing:

Archomeda:
Hmm, it really is a tough decision (at least for me) over something that's this small. Both ways are reasonable and have valid arguments. In the end, it doesn't really matter to me how it's encoded, I was just adding my thoughts :smile:

Anyway, I guess this whole PR is drifting off towards that.
About the items themselves, either I'm missing totally obvious, but what if an item in the bank (or inventory) is transmuted? Currently, I'm not seeing an example for that. Would it be best to have a `skin` entry when it's transmuted (just like `/v2/items` contains a `default_skin` entry per item)?

lye:
Okay, updated the example to contain items that have skins and infusions. Also switched empty bank slots to be indicated by `null`, I think it makes more sense that way.

For non-empty items, the item encoding is going to be standard across all endpoints that return items (e.g., the bank/character inventories/material storage, etc).

StevenLiekens:
Does it make sense to describe empty slots in the output if you also describe the slot number for each non-empty slot?

In other words: we only get objects for slots 1, 2, 3, 4, 10 and 11. From that, it's easy to deduce that slots 0, 3, 5, 6, 7, 8, 9 and 12 to 30 are empty. Does it make sense to describe them?

tivac:
@StevenLiekens It's a nice convenience for people who are consuming the APIs directly on the front-end. It makes turning the API response into UI simpler & doesn't cost us much on our side since we gzip results.

JoluMarti:
Bound: Check if a item is soulbound or account bound.
```
{
		id: 48087,
		slot: 2,
		count: 1,
		bound: "Soulbound",
		skin: 89
},
```

zeeZ:
Better yet: Include some form of reference to the character/account the item is bound to? That should be done for all places where bound items could pop up, so I'm not sure if this here is the place to have lengthy debates about that.

darthmaim:
@zeeZ from the old original PR #3:
>>Can the bank schema include info re: to which character an item is soulbound?

>It can, but that feature probably won't make it for the initial release, so I'm going to leave it out for now.

PR #13, opened 2015-02-27
Initial pass at /v2/materials and /v2/account/materials

lye:
These endpoints expose the items stored in an account's bank in the material storage. It provides an additional unauthenticated endpoint that describes which items can be stored and the category headings. These changes are largely based (and forked off from) the discussion in #3.


Ruhrpottpatriot:
Seems reasonable. I will definitely look into it in a few hours after I got some well deserved sleep. Unauthenticated APIs are always good (more for me to implement in GW2.NET) :)

PR #16, opened 2015-03-04
Initial pass at /v2/characters

lye:
`/v2/characters` is an authenticated bulk-expanded endpoint that's intended to dump information about an account's characters. Currently, it requires both `account` and `characters` scopes, and provides additional fields if you have the `inventories` scope. Check the examples -- it's pretty straightforward.

The `inventory` and `bags` fields work in the same manner as `/v2/account/bank` -- you get an array where each element represents a slot. If the slot is empty, there's `null` at that index. The items are encoded in the same way as `/v2/account/bank`, makes it a bit easier. I've included the `size` field for the bags since I needed it internally and figured it'd be useful for laying out the items (without the overhead of including the entire item blob for each bag).

`equipment` also uses the same "item stack" encoding, with the exception that it has a "slot" field thrown in. See the concerns section for concerns on this.

`guild` is the GUID of the currently represented guild (and can be used against `/v1/guild_details.json?guild_id=X`), or null if no guild is represented.

### My current concerns:

 * I'm not entirely sold on dumping all this information (specifically, the equipment/inventory) into the same endpoint, since we split it off the `/v2/account` endpoint. Splitting it off here is a bit more involved since we're dealing with a bulk-expanded endpoint, but it's workable. I'm not sure if there's added benefit to doing this.
 * Should `guild` be hidden behind a scope? My thinking is no, since it's clearly visible in-game and the fewer scope complications the better.
 * The `equipment.slot` stuff is looking really gross to me today. There's no way to determine which weapon is main/offhand. What's a better way to encode this? I'm thinking maybe make the slot field an integer where  multiples of the same type can be equipped (rings, accessories and weapons). So you might have weapons in slots `0, 2, 3` (where the first one is either two-handed or without an offhand).

Thoughts?

### DIFF
--- a/v2/characters/characters.js
+++ b/v2/characters/characters.js
@@ -39,86 +39,71 @@
 	profession: "Thief",
 	level: 80,
 	guild: "1F5F70AA-1DB6-E411-A2C4-00224D566B58",
-	equipment: [
-		{
-			id: 69091,
-			slot: "HelmAquatic"
+	equipment: {
+		aquatic_helm: {
+			id: 69091
 		},
-		{
+		backpack: {
 			id: 49371,
-			slot: "Backpack",
 			infusions: [
 				49453
 			]
 		},
-		{
+		coat: {
 			id: 48085,
-			slot: "Coat",
 			infusions: [
 				37131
 			]
 		},
-		{
-			id: 48084,
-			slot: "Boots"
+		boots: {
+			id: 48084
 		},
-		{
-			id: 48086,
-			slot: "Gloves"
+		gloves: {
+			id: 48086
 		},
-		{
-			id: 48088,
-			slot: "Leggings"
+		leggings: {
+			id: 48088
 		},
-		{
-			id: 48089,
-			slot: "Shoulders"
+		shoulders: {
+			id: 48089
 		},
-		{
-			id: 39233,
-			slot: "Accessory"
+		accessory1: {
+			id: 39233
 		},
-		{
-			id: 39232,
-			slot: "Accessory"
+		accessory2: {
+			id: 39232
 		},
-		{
-			id: 37075,
-			slot: "Ring"
+		ring1: {
+			id: 37075
 		},
-		{
-			id: 37086,
-			slot: "Ring"
+		ring2: {
+			id: 37086
 		},
-		{
-			id: 39308,
-			slot: "Amulet"
+		amulet: {
+			id: 39308
 		},
-		{
-			id: 46771,
-			slot: "Weapon"
+		weapon_mainhand_primary: {
+			id: 46771
 		},
-		{
-			id: 46760,
-			slot: "Weapon"
+		weapon_offhand_primary: null,
+		weapon_mainhand_secondary: {
+			id: 46760
 		},
-		{
-			id: 46767,
-			slot: "Weapon"
+		weapon_offhand_secondary: {
+			id: 46767
 		},
-		{
-			id: 23029,
-			slot: "Sickle"
+		aquatic_weapon_primary: null,
+		aquatic_weapon_secondary: null,
+		sickle: {
+			id: 23029
 		},
-		{
-			id: 23030,
-			slot: "Axe"
+		axe: {
+			id: 23030
 		},
-		{
-			id: 23031,
-			slot: "Pick"
+		pick: {
+			id: 23031
 		}
-	],
+	},
 	inventory: [
 		null,
 		null,
--- a/v2/characters/characters.js
+++ b/v2/characters/characters.js
@@ -1,0 +1,228 @@
+// GET /v2/characters
+// Authorization: Bearer token123
+// Scopes: account, characters
+[ "Hello" ]
+
+// GET /v2/characters/Hello
+// Authorization: Bearer token123
+// Scopes: account, characters
+{
+	name: "Hello",
+	race: "Human",
+	gender: "Female",
+	profession: "Thief",
+	level: 80,
+	guild: "1F5F70AA-1DB6-E411-A2C4-00224D566B58"
+}
+
+// GET /v2/characters?page=0
+// Authorization: Bearer token123
+// Scopes: account, characters
+[
+	{
+		name: "Hello",
+		race: "Human",
+		gender: "Female",
+		profession: "Thief",
+		level: 80,
+		guild: "1F5F70AA-1DB6-E411-A2C4-00224D566B58"
+	}
+]
+
+// GET /v2/characters/Hello
+// Authorization: Bearer token123
+// Scopes: account, characters, inventories
+{
+	name: "Hello",
+	race: "Human",
+	gender: "Female",
+	profession: "Thief",
+	level: 80,
+	guild: "1F5F70AA-1DB6-E411-A2C4-00224D566B58",
+	equipment: [
+		{
+			id: 69091,
+			slot: "HelmAquatic"
+		},
+		{
+			id: 49371,
+			slot: "Backpack",
+			infusions: [
+				49453
+			]
+		},
+		{
+			id: 48085,
+			slot: "Coat",
+			infusions: [
+				37131
+			]
+		},
+		{
+			id: 48084,
+			slot: "Boots"
+		},
+		{
+			id: 48086,
+			slot: "Gloves"
+		},
+		{
+			id: 48088,
+			slot: "Leggings"
+		},
+		{
+			id: 48089,
+			slot: "Shoulders"
+		},
+		{
+			id: 39233,
+			slot: "Accessory"
+		},
+		{
+			id: 39232,
+			slot: "Accessory"
+		},
+		{
+			id: 37075,
+			slot: "Ring"
+		},
+		{
+			id: 37086,
+			slot: "Ring"
+		},
+		{
+			id: 39308,
+			slot: "Amulet"
+		},
+		{
+			id: 46771,
+			slot: "Weapon"
+		},
+		{
+			id: 46760,
+			slot: "Weapon"
+		},
+		{
+			id: 46767,
+			slot: "Weapon"
+		},
+		{
+			id: 23029,
+			slot: "Sickle"
+		},
+		{
+			id: 23030,
+			slot: "Axe"
+		},
+		{
+			id: 23031,
+			slot: "Pick"
+		}
+	],
+	inventory: [
+		null,
+		null,
+		null,
+		null,
+		{
+			id: 12138,
+			count: 217
+		},
+		null,
+		{
+			id: 626,
+			count: 1,
+			upgrades: [
+				24797
+			]
+		},
+		null,
+		null,
+		{
+			id: 19976,
+			count: 1
+		},
+		{
+			id: 67134,
+			count: 1
+		},
+		{
+			id: 8934,
+			count: 1
+		},
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
+		null,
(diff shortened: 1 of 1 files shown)


aRestless:
> Should guild be hidden behind a scope?

As far as I've seen we haven't talked about scopes for guilds at all yet, have we? My only concern with guilds would be, that you might want to _only_ expose that you're owner of account X and that this account is member of guild Y (for example as authentication for guild forums), but nothing else. So there might be use cases for a `guild` scope, but not necessarily in this case.

There also might be use cases where you want to grant an `equipment` scope (from gearchecks to fashion viewers), but not the one for `inventory` - or would that be oversophisticated?

darthmaim:
I feel like we should split inventory off of the character endpoint into an extra `/v2/characters/{character}/inventory` endpoint. 

What if my app has the inventory scope for a rarely used feature that gets only ever activated by the user, but wants to crawl the character endpoint (level/guild/...) regularly to show up to date statistics about a character. I don't need/want to request all the inventory stuff once every hour for potentially thousands of characters I want to update.

For the equipment, since we have a fixed number of slots that never change, maybe something like this:
```js
{
  // [...]
  equipment: {
    boots: { id: 48084 },
    gloves: { id: 48086 },
    backpack: { id: 49371, infusions: [ 49453 ] },
    weapon_mainhand_primary: { id: 46771 },
    weapon_offhand_primary: null,
    weapon_mainhand_secondary: { id: 46760 },
    weapon_offhand_secondary: { id: 46767 },
    ring1: { id: 37075 },
    ring1: { id: 37086 },
    // [...]
  }
}
```
This kinda removes the similarities to the bank/inventory endpoints by changing the array to an object, but the display of the equipped items probably has to be totally different anyway. And we can still loop over it and get what was `slot` in your proposal as key in the loop.

This would have the upside of easy access to specific slots, for example, to get the currently equipped weapons to display available skills in a skill calculator with `/v2/skills`.

Also, wouldn't it make more sense to `GET /v2/characters/{some_character_guid}` instead of `GET /v2/characters/Hello`? Then nothing would break in the rare case of a rename ([Name Change Contract](http://wiki.guildwars2.com/wiki/Name_Change_Contract), ...). We request items, guilds, skills, ... all by (GU)ID, and we wouldn't have to deal with encoding in the URL. (Yes I read the "*Characters do not have an immutable identifier that we can expose*" on the forum ([source](https://forum-en.guildwars2.com/forum/community/api/Launching-v2-account-w-Authentication/4830070)), but a immutable identifier would really help us.)

lye:
> So there might be use cases for a guild scope, but not necessarily in this case.

I feel roughly the same way. I want to reserve the `guild` scope to grant access to MOTD/roster/history, I think. I don't think the account/character guilds need to be protected by additional scopes.

> Equipment slots as object keys.

I kind of like this design, it removes the gross special-casing that I added. I'll change my local implementation to use that then update the pull request with it.

> Also, wouldn't it make more sense to GET /v2/characters/{some_character_guid}

It would, but character GUIDs are strictly secret. We could theoretically pass them through a one-way transform (e.g., a hash function), but I think it would take significant backend work to plumb that through everything. The character name is at least unique; the mutability is just a constraint I think we'll have to live with.

I'll spend some more time looking into it though. It's a really horrible constraint to live with.

lye:
Ah, missed this --

> I feel like we should split inventory off of the character endpoint into an extra /v2/characters/{character}/inventory endpoint.

I'm kind of feeling that too, but want to think about it some more. Do we want to split equipment from the inventory? Should the bags also be included in `v2/character/{name}/inventory`? It almost makes sense to just return all of `{inventory,bags,equipment}` there.

darthmaim:
I think bags should definitely be included in the inventory endpoint, it doesn't make much sense to get the bags without the inventory, and while some may just want the inventory and don't need the bags, the response is so small it doesn't hurt to include it.
I'm not too sure about the equipment endpoint, I guess whatever works best for you.

tivac:
@lye and I talked IRL & agreed taht equipment should be broken out. as `/v2/character/{name}/equipment`.

After chatting about it we were thinking about rolling inventory up under bags, since the two are pretty tightly linked:

```js
// /v2/character/{name}/inventory
{
    bags: [
        {
            id: 8941,
            size: 4
            items : [
                {
                    id: 12138,
                    count: 217
                },
                null,
                null,
                null
            ]
        },
        null,
        {
            id: 8932,
            size: 20,
            items: [
                null,
                null,
                // ...
            ]
        },
        {
            id: 48715,
            size: 20,
            items : [
                // ...
            ]
        },
        {
            id: 48715,
            size: 20,
            items : [
                // ...
            ]
        }
    ]
}
```

darthmaim:
Sounds good to me! :+1:

Remfin:
Is there a technical reason for different endpoints? API responses can't vary based on query string or would only respond to scopes for the token regardless of what is asked for? Because if you're refreshing an account's characters completely, separate endpoints mean what would be 1 call for an 8-character account is now 17.

I know you have to draw the line somewhere (why not just let the account API return everything! :)), but it's hard to imagine a functional app that isn't going to be equipment, or equipment+inventory (after all, account-bound equipment that's being worn is potential inventory if you're an inventory-tracking app).

As for equipment, not shown or not available?
Current skin
Dye(s) (I think this is actually on the character, not really the item)
Prefix/suffix (legendaries)

lye:
> Is there a technical reason for different endpoints?

No, the backend that we talk to returns all the data for a single request. Just tryin' to do what makes the most sense.

> Current skin

That should be shown, just don't have any transmuted items on "Hello" right now.

> Dye(s), Prefix/suffix

Won't be in the initial release ;_;

Remfin:
Put me down for making them one API with a query string to pick and choose or something. Leave the decision up to the API caller and avoid imposing extra HTTP calls just for the sake of things looking more REST-y/object-y. Also lets you benefit a lot more from the gzip compression.

What happens with inventory overflow? :)

aRestless:
@tivac @lye Looks way better now, especially the bags-first hierarchy. Dyes would be great and although they won't be in the initial release, this might already be the right time to think about how they'll be represented eventually.

Archomeda:
It looks good. I really like the way how the inventory is being handled. Also, having them separated is fine with me. In case you want to extend the characters with more character bound data, you can just add another endpoint without cluttering the main one.

Just a couple of notes (which may or may not have been answered somewhere yet):
- Is it possible to include the creation date of a character? Or even better yet, the exact day and time in ISO 8601 if it's available. I think this might also help applications to identify a character if the name has been changed by iterating all characters (I really doubt if people manage to create two characters on the exact time)
- Is including the selected skills and traits possible?
- Adding from the previous one, what about the upcoming specializations?
- Other (less important stuff): skill points, upcoming masteries (although this might be suitable for /v2/account, don't really know how they unlock/work yet), current active outfit/mini/finisher/mail carrier/etc, probably more that I can't think of right now

I assume that talking about HoT features in the APIs might be too soon though, since there's no known release date of HoT. But they are just some thoughts.

lye:
> What happens with inventory overflow? :)

Inventory overflow is considered a completely separate "inventory", e.g., there's one list of items for the "bags inventory", another for the "bags' contents inventory" and a third for the "overflow inventory". We're only currently pulling data from the first two, so the overflow inventory won't currently show up at all.

> Dyes would be great and although they won't be in the initial release, this might already be the right time to think about how they'll be represented eventually.

I was just gonna toss the dye color IDs into the itemstack object

```json
{
    id : 134,
    count : 1,
    dyes : [ 23, 456, 284 ]
}
```

> Is it possible to include the creation date of a character? Or even better yet, the exact day and time in ISO 8601 if it's available.

That seems pretty reasonable, I'll see if I can dig up where that's stored.

> Is including the selected skills and traits possible?

Yes, but I want to put that off until `/v2/skills` and `/v2/traits` are finished. Not too much you can do with a list of IDs.

> Other (less important stuff)

Those all seem fairly reasonable. None of that's currently exposed to us, but I'll try to root out where that's stored. The mail carrier/finisher/etc are probably not going to be useful without having an endpoint to expose their data, but SP/outfit/minis should be a bit easier.

> I assume that talking about HoT features in the APIs might be too soon though, since there's no known release date of HoT. 

Yeah, I don't know enough about HoT features to work meaningfully on the endpoints. There's plenty of stuff from the core game to expose.

Remfin:
Some numbers for my argument to make it a unified API (if you also want the single APIs, doesn't matter to me :-))

Random Name, Profession, Gender, Race, Level, random GUID for guild, randomized equipment for all slots serialized in the proposed format (pretty printed, long descriptive keys for slots), GZip.Fastest

1 character, no equipment:
Uncompressed: 231
Compressed: 181

1 character, equipment:
Uncompressed: 1250
Compressed: 345

8 characters, equipment:
Uncompressed: 11211
Compressed: 1047

A _single_ SSL handshake (done for every single API call) is at least 3x larger than my "large" call, let alone the time overhead.

If the data back-end supports it, less connections is so much better :)

codemasher:
HTTP2 soon&trade;

https://mattwilcox.net/web-development/http2-for-front-end-web-developers

tivac:
@codemasher :smirk:

Chatted w/ @lye for a bit. We think the best plan is for `/v2/characters` to return all information at the top-level, and then you can drill down via `/v2/characters/<name>/inventory` or `/v2/characters/<name>/equipment`.

koenvu:
I was wondering whether the gathering tools will have a key like `uses_left`. A sample use-case for this would be to notify users they might need to grab some new tools if they are going on a gathering spree.

Current:

```
...
		sickle: {
			id: 23029
		},
		axe: {
			id: 23030
		},
		pick: {
			id: 23031
		}
...
```

Proposed:

```
		sickle: {
			id: 23029,
			uses_left: 23
		},
		axe: {
			id: 23030,
			uses_left: 11
		},
		pick: {
			id: 23031,
			uses_left: 39
		}
```

The only question would be: how to handle infinite gathering tools. Three options I see:

1. Add another key (boolean): `is_infinite`. If this key is`true`, the `uses_left` key would be meaningless. Alternatively the developer could maybe deduce this from the items API? Not sure which flags are available there for items.
2. `uses_left` would contain `null`, indicating it is not applicable for this tool.
3. `uses_left` would contain (int) 0, since all other gathering tools are automatically destroyed when they have no uses left. Alternatively it could be -1, but I find that even more than an ugly hack than this option already is.

tivac:
I'd prefer `uses`, and either `null` or omitting the field entirely for infinite gathering tools. Curious what data is actually available and will ask @lye about it when he's in.

lye:
Tool uses isn't currently pulled out of the character data, but it's something I can look into adding.

mattdeboard:
Please don't omit fields from the data because the attribute isn't applicable! Setting to `null` is just as easy on server-side, and gives clients predictability through "type safety" (such as it is in JS).

Barring that please provide metadata describing the shape of the respose or custom content type headers so clients don't have to do gloopy `if (!!response.uses) {` guesswork all over the place.

But homogenous data representation for a given resource is far preferable, especially when the choice is between omission and `null`. (Though `-1` is probably a better value for infinite-use gathering items than null, for the sake of aforementioned type consistency).

charlieman:
I'd really like to see the guild bank log in the API. I'd be great for my lotto application to be able to parse it and know about the gold people has deposited/withdrawn.

KevinVitale:
Personally, it'd be interested to get the characters' current/recent in-game location. Thoughts?

mstampfli:
Just wondering, is it possible to add not only the current % of the map Complition, but also the exact pois, vistas hearts which a certain character has done/not done? I'm not sure how to display them, but working out on a good way.

lye:
Sorry for taking awhile to get back.

> I'd really like to see the guild bank log in the API.

Noted, though this isn't the right PR for that :P

> Personally, it'd be interesting to get the characters' current/recent in-game location. Thoughts?

Definitely something to look into. I'm not sure what the TTL will be on that data, or how much delay applications can tolerate. Our current data source for character data can be pretty stale but I can probably get the current location from elsewhere (since it seems to be fairly up-to-date in both the guild and contacts panels).

> Just wondering, is it possible to add not only the current % of the map Complition, but also the exact pois, vistas hearts which a certain character has done/not done? 

Will probably not be in the first release (since I have no idea how/where that information is currently stored), and the per-POI bits will likely be on a separate endpoint since it's fairly niche data. Don't really have an ETA for it.

codemasher:
As per [this thread](https://forum-en.guildwars2.com/forum/community/api/Played-time-age-api/first#post5138470) again: would it be possible to add a characters age/creation date? (same for /v2/account)

lye:
Yeah; I'll open separate PRs later today for them though. Makes it a bit easier to track.

PR #25, opened 2015-05-19
Proposal for v2/account/mails

Teranas:
Hi there,

and again a new proposal. This time for an endpoint that enables us to request ingame mails.

I hope it's possible to provide us this data.

Please let me know what you think.

~ Teranas


lye:
The items need to include quantity as well. The other proposals are using the format -

    { "id" : 1234
    , "count" : 42
    }

Teranas:
Just added the 'count' property and transformed items to objects.

StrategistXIV:
This looks good! This might be a silly question, but is there a possibility this could also be extended so that we could send mail with this API? I would change the endpoint into v2/account/mail (without the s) though.

HarveyEV:
Agreed. Ability to send mail to game through the API would be incredibly useful.

Coffee4cr:
As long as it's restricted to guild members

StrategistXIV:
If there is a possibility (Which I think would be hard to implement) then we need to make sure that bots cannot use this API to spam. I think we would have a few possibilities:

At one point it might be 'easy' to put a limit on the about of mails you can send for some amount of time, but this isn't fool proof. I heard stories about people owning multiple bots, which could potentially all be used for this case together.

Another thing we might do is limiting the sender to only send to specific persons:
- Limiting to replying to people that sended you a message
- Limiting to sending people in your friend list
- Limiting to sending people in the guild.

A combination of the above could also be possible. I do think that we would face the same issue if we would implement a Chat API. Which would allow us to see things on the chat. Personally I think a Chat API could really enhance the community (especially Guilds) since people would also talk to each other outside the times they can play GW2, which would strengthen the bond of the guild.

darthmaim:
I highly doubt that there will be a way to send mails via the API, but there has been discussion about a chat API in the past:

> Out-of-game guild chat is definitely something I want to have ~@lye ([source](https://forum-en.guildwars2.com/forum/community/api/API-CDI-2015/4813732))

StrategistXIV:
I created a PR for the chat API darthmaim: #43

psykzz:
Mail age needs to also be consider, an epoch timestamp would be ideal.

lye:
Since this was bumped, I should probably give an update with some additional givens:

 * We only store the mail that's displayed in the UI (at least, for real-time retrieval) -- that means neither historical mails, nor any mail that is not received (e.g., you can't see mail that you've sent).
 * We cannot send mail via the API -- though it's technically possible, that subsystem (for writing to character data from an unauthoritative source) scares me and I have a feeling something would go horribly wrong.
 * The API can't mark mail as read/unread, and cannot delete mail messages.

Given these constraints, the usefulness of a mail endpoint seems really limited to me.

lye:
Okay, so I ended up writing a bit that pulls data from the mail queue, so the API can view the mail before it's actually delivered (delivery requires login).  With this, the API can show the 10 messages displayed in the UI, and the first 90 messages in the delivery queue (assuming the delivery queue can get that deep -- not entirely sure).

The read-only restriction isn't ever going to change, though. I think the ability to read up to 100 messages (without requiring login to read new messages) greatly changes the usefulness of the endpoint, though.

I forgot about this PR and ended up opening another one, so this is superseded by #518.

PR #28, opened 2015-05-28
Proposal for adding map type to /v2/continents and /v2/maps

Archomeda:
I came up with this idea about a month ago on the forums, but got no response. So I decided to elaborate it further with this PR.

This addition makes our lives easier to separate the various maps ("normal" pve, pvp, wvw, dungeons, story, etc.) without defining a list on our own (that might be changing with updates too!).

Currently, I've compiled the following list of map types:
- pve (for normal pve maps)
- pvp
- wvw
- personal_story
- living_story
- dungeon_story
- dungeon_explorable

I'm pretty sure I missed some. Let me know if you like the idea and if it's actually possible to include it. You do have a way of identifying different maps internally, right? :grinning:

On a side note, I added /v2/maps as well with the current implementation and my addition.

### DIFF
--- a/v2/floors/map.js
+++ b/v2/floors/map.js
@@ -1,6 +1,7 @@
 // GET /v2/continents/1/floors/1/regions/1/maps/26
 {
     "name" : "Dredgehaunt Cliffs",
+    "type" : "pve",
     "min_level" : 40,
     "max_level" : 50,
     "default_floor" : 1,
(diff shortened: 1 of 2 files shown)


StevenLiekens:
I'd like to add to this that the Mumble Link context object has a numeric field named `mapType`, and that it would be helpful if we could correlate these values with the values returned by the API.

Archomeda:
Hmm, you made me remember a project I used to work on, but stopped working on it after I couldn't solve some lag issues that started over time. [I tested some of the maps with that field](https://github.com/Archomeda/OBS-GW2-Plugin/blob/master/LuaScripts/ScriptVariables/GameModeName.lua). The numbering is a bit weird. Plus I'm missing some in between, if they're used at all.

lye:
> You do have a way of identifying different maps internally, right?

Ah, umm, I hope so? I haven't really looked at the maps stuff much.

> the Mumble Link context object has a numeric field named mapType

Huh, I'll dig up that enumeration tomorrow morning and, if they're meaningful values, put 'em into the mumble.md file in this repo. Not sure how I missed that when I originally wrote that file.

Archomeda:
Closing this as it's been superseded by #333.

PR #33, opened 2015-06-12
Add created/age/deaths to account/character endpoints

lye:
The oft-requested feature; it's pretty straightforward. The new values will be accessible using the current scopes ("account" and "characters", respectfully). `age` is measured in seconds and the dates are ISO8601-formatted.


lye:
`created` should now be returned by `/v2/account`; the additional fields in `/v2/characters` should start showing up tomorrow after the patch.

PR #34, opened 2015-06-13
Proposal to add delivery items/gold to /v2/commerce/transactions/

risingashes:
Currently we have current and historical data, but nothing to indicate whether the user has picked up the items or not.

My proposal adds this data in a format similar to current and historical.

### DIFF
--- a/v2/characters/characters.js
+++ b/v2/characters/characters.js
@@ -1,0 +1,213 @@
+// GET /v2/characters
+// Authorization: Bearer token123
+// Scopes: account, characters
+[ "Hello" ]
+
+// GET /v2/characters/Hello
+// Authorization: Bearer token123
+// Scopes: account, characters
+{
+	name: "Hello",
+	race: "Human",
+	gender: "Female",
+	profession: "Thief",
+	level: 80,
+	guild: "1F5F70AA-1DB6-E411-A2C4-00224D566B58"
+}
+
+// GET /v2/characters?page=0
+// Authorization: Bearer token123
+// Scopes: account, characters
+[
+	{
+		name: "Hello",
+		race: "Human",
+		gender: "Female",
+		profession: "Thief",
+		level: 80,
+		guild: "1F5F70AA-1DB6-E411-A2C4-00224D566B58"
+	}
+]
+
+// GET /v2/characters/Hello
+// Authorization: Bearer token123
+// Scopes: account, characters, inventories
+{
+	name: "Hello",
+	race: "Human",
+	gender: "Female",
+	profession: "Thief",
+	level: 80,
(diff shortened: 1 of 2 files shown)


lye:
Honestly, I'd much rather these be a separate endpoint -- the data for the transactions endpoint comes from a completely separate source than the delivery box data. The endpoint is already jumping through some hoops to re-page from one paginated endpoint to another, and there's no clean way to insert additional records.

`/v2/commerce/delivery` might make sense, not sure if I'm sold on the diction though.

StrategistXIV:
I like the idea of this API! I do think that can be a little confusion though. What about something like this:

/v2/commerce/transactions/delivered/

            {
                 gold: 32144  // This is 0 when no gold is there to be picked up.
                 items:
                 [
                      {
                      id: 1997,
                      item_id: 19699,
                      quantity: 20,
                      transaction_date: "2014-12-15T14:43:36+00:00"
                      },
                      {
                      id: 1999,
                      item_id: 19699,
                      quantity: 20,
                      transaction_date: "2014-12-15T14:43:35+00:00"
                      },
                      ...
                 ]
            }

Then we wouldn't need an extra endpoint only for gold. I wonder what you guys think about this.

asteingass:
Hello everyone, any news about this topic?

lye:
This is technically feasible with the following caveats: 

 * It won't be real-time. It can be delayed up to 10 minutes.
 * The data won't be updated unless the user is currently logged in.

Those limitations IMO kill the utility of the endpoint, so it's pretty low-priority. Let me know if those constraints aren't as bad as they sound.

queicherius:
I think it not being real time or not having the transactions is completely fine. The non-updating kills when people are not logged in, however removes nearly all use-cases I had in mind for a feature like this (aka "I am away, and I want to see what is in there / I want to get notifications").

---

Just gonna link #321 here so the cross-reference exists

asteingass:
Hi, thanks for the fast answere.
Currently a friend of mine and I are working on some stuff, and we liked the idea of seeing your current delivery box.
The delay wouldn't be a big point. 
And as @queicherius already mentioned, we also had this idea of "I am away, and I want to see what is in there". So if there is no other way, it's not that important anymore for us.

Thanks
Andreas

Darqam:
If you mean login to the game, no. The API can only read info from the game, it has no capability of writing to the game in any sort of way.

lye:
Finally managed to work out the kinks in this.

Interestingly, we don't actually record the id/transaction time with the coin/items that are awaiting pickup. There's zero correlation with `/v2/commerce/transactions`, which is a bit unfortunate. I've reworked the output format around to compensate for this.

#514 supersedes this.

PR #41, opened 2015-06-22
Initial pass at /v2/pvp/stats

lye:
`/v2/pvp/stats` is an authenticated endpoint that's intended to provide aggregated pvp statistics for an account. These are roughly the metrics that are displayed in the in-game PvP panel.

I'm not entirely sold on how `professions`/`ladders` are laid out -- it may make more sense to structure the data differently. I'm also not sure what `forfeits`/`byes` are in the context of PvP -- I suspect they're to do with one team not accepting a popped queue (if that's the case, then those fields will need to be renamed).

AFAIK we don't track per-ladder profession data, that's probably something that should be added too.

### DIFF
--- a/v2/pvp/stats.js
+++ b/v2/pvp/stats.js
@@ -3,6 +3,7 @@
 
 [
 	{
+		pvp_rank  : 2,
 		aggregate : {
 			wins       : 1,
 			losses     : 2,
--- a/v2/pvp/stats.js
+++ b/v2/pvp/stats.js
@@ -1,0 +1,46 @@
+// GET /v2/pvp/stats?access_token=token
+// Scopes: account, pvp
+
+[
+	{
+		aggregate : {
+			wins       : 1,
+			losses     : 2,
+			desertions : 0,
+			byes       : 2,
+			forfeits   : 0
+		},
+		professions : {
+			mesmer : {
+				wins       : 0,
+				losses     : 1,
+				desertions : 0,
+				byes       : 1,
+				forfeits   : 0
+			},
+			thief: {
+				wins       : 1,
+				losses     : 1,
+				desertions : 0,
+				byes       : 1,
+				forfeits   : 0
+			}
+		},
+		ladders : {
+			unranked : {
+				wins       : 0,
+				losses     : 1,
+				desertions : 0,
+				byes       : 0,
+				forfeits   : 0
+			},
+			ranked : {
+				wins       : 2,
+				losses     : 1,
+				desertions : 0,
+				byes       : 2,
+				forfeits   : 0
+			}
+		}
+	}
+]


StrategistXIV:
I think this will also be a great API! Personally I also don't know what forfeits or byes are. Maybe the difference between the two is disconnecting vs being afk? A few things we might need to look into:

- Are we going to provide a field with "Favorite Profession", or do you think we can calculate that from the stats?
- Favorite Map might be something fun to add. (If it's worth it)
- "Amount of top stats awarded" field, I am not sure how many people value this field, but we might want to add it for completeness.
- Maybe we should add the sPvP rank of the account, and maybe even the experience points? (So applications can be made that can calculate how many battles you still need to play for your next rank)

When first viewing the professions layout, it felt strange to me too. Right now we have no way to see how many unranked/ranked matches are played with a specific profession. Maybe it's best to include the ladder data inside the profession data? (Even though that does look a bit awkward, it might be the only way to achieve this.)

Like this:

    [
        professions : {
            mesmer : {
                unranked : {
                    wins       : 0,
                    losses     : 1,
                    desertions : 0,
                    byes       : 0,
                    forfeits   : 0
                 },
                 ranked : {
                      ...
                 }
            }
        }
     ]

This way we can count up all the matches for each profession (shouldn't be too hard in most programming language's I think) to get the amount of matches played in total by all professions. 

The only thing I didn't really understand was what you ment by aggregate.

I hope I can be of any help discussing this.

lye:
> Are we going to provide a field with "Favorite Profession", or do you think we can calculate that from the stats?

Should be able to compute that from the stats -- AFAIK it's the profession with the most completed games (e.g., `wins + losses`). Favorite map might be a bit trickier -- that's definitely stored somewhere, but I'm not yet sure where. Should definitely be exposed though.

> Maybe we should add the sPvP rank of the account, and maybe even the experience points?

Ooh, that's a really good point. I've been meaning to find a place for that.

> The only thing I didn't really understand was what you ment by aggregate.

I'd have to double-check, but I believe the current "aggregate" is basically data from all matches (including unrated matches; e.g., hotjoin). Probably can find a better term.

I'm not yet that familiar with the backend components that are powering my current implementation; I'm mostly spitting out what the backend currently provides. I'm not sure how much wiggle room there is (e.g., whether we track per-ladder profession stats), but it's something I'll look into. I'm pretty sure it's displayed on the PvP panel though, so it should be in there somewhere.

HarveyEV:
This all looks great. I don't see any RFC for leaderboard endpoint so I thought best to post my question here here. 

Also, are there any plans or discussions about how guild teams will be reflected? As they will purportedly have their own leader boards, how would that affect the structure (if there is one)? Is there some overlap possibly with the guild endpoint to display spvp teams within a guild?

lye:
> Also, are there any plans or discussions about how guild teams will be reflected?

I actually have no idea how guild teams will work; the following is idle speculation. Ideally they work just like normal matches and we can just stuff an extra field or two in the game objects which indicate which color was which guild team and such.

It would be ideal to just include a "guild team id" for each color (allowing it to be `null`/`undefined`) and then having a separate endpoint which provides the metadata for that team, but I'm not sure how mutable the teams will be so that approach might end up with a mismatch between the current guild team state and the state at the time of the match.

And there's a whole bunch of other possible edge-cases, like, if you're in an 8v8 custom arena and your team is made up of 2x 4-man guild teams. Not really sure if that's possible or how we can express that succinctly. 

/ramble

Eearslya:
Couple quick questions as someone who doesn't play PvP often.

What exactly is the difference between `desertions`, `byes`, and `forfeits`?

And what is the `none` subcategory under the `ladders` object?

lye:
> What exactly is the difference between desertions, byes, and forfeits?

Gonna have to get back to you on that; I'm not entirely sure myself.

> And what is the none subcategory under the ladders object?

Ladders are weird; either you've got some garbage data in your account (possible!) or there was a ladder named "none". Might be data from an era around launch or something, not really sure. Haven't rooted around in the actual game server code (which is the upstream of the upstream of the data source used for this endpoint).

@tynril might know more, not sure.

PR #42, opened 2015-06-24
Initial pass at /v2/specializations

lye:
Just another bulk-expanded endpoint that dumps specialization data. Works the same as all the other ones. The `minor_traits` and `major_traits` contain trait IDs that may be resolved against `/v2/traits` which I haven't written up a PR for since it's a bit more involved (and currently a bit buggy).

### DIFF
--- a/v2/specializations/specializations.js
+++ b/v2/specializations/specializations.js
@@ -1,0 +1,84 @@
+// Without parameters, returns a list of available IDs.
+// GET /v2/specializations
+
+[ 1, 2, 3, 4, ... ]
+
+// IDs may be requested individually to resolve objects.
+// GET /v2/specializations/4
+
+{
+    id: 4,
+    name: "Inventions",
+    profession: "Engineer",
+    elite: false,
+    minor_traits: [
+        518,
+        508,
+        519
+    ],
+    major_traits: [
+        394,
+        2037,
+        507,
+        1707,
+        2023,
+        445,
+        472,
+        1709,
+        2024
+    ],
+    icon: "http://render.local.ncplatform.net/spawned/RenderSrv/file/BD349FDE30ECDCD6D3113A2FCEC414D3E5553F9E/1954407.png"
+}
+
+// Or in bulk, either by ID or paginated.
+// GET /v2/specializations?ids=4,5
+// GET /v2/specializations?page=0&page_size=5
+
+[
+    {
+        id: 4,
+        name: "Inventions",
+        profession: "Engineer",
+        elite: false,
+        minor_traits: [
+            518,
+            508,
+            519
+        ],
+        major_traits: [
+            394,
+            2037,
+            507,
+            1707,
+            2023,
+            445,
+            472,
+            1709,
+            2024
+        ],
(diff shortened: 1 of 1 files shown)


StrategistXIV:
This looks interesting! I am sure this will become a useful API once the Trait API is also fully released. A few comments that came to mind:

- Maybe Elite: true/false should be changed to. Type: 'Elite' just in case some other type of traitline comes out? (I doubt it thought, but just in case)

- Maybe we should extract Profession out each entry and group all specializations per profession? I think this API will be most used for build calculators, so when this request is made it is probably interested in only one profession. Instead of having to filter for each profession they can just take the profession node and parse the Specializations in there? It's just an idea though, I think what you have right now is also already appealing. I ment something like this:

    [
        Engineer: 
        [
             {
                 id: 4,
                 name: "Inventions",
                 ....
             },
             {
                 id: 5,
                 name: "Alchemy",
                 ....
             },
             ....
        ],
        ...
    ]

- Another small thing we might to consider adding, is how many Hero Points are required to unlock a certain trait. If we keep the ordering of traits the same as the order of unlocking, then we would only have to add the amount of Hero Points needed to provide this data.

That's all that came to mind!

lye:
> Maybe Elite: true/false should be changed to. Type: 'Elite'

@tivac, I like `true`/`false` but given the rest of our data `Elite` is more idiomatic. Thoughts?

> Maybe we should extract Profession out each entry and group all specializations per profession?

Eh, there's few enough specializations that I think it makes the most sense for the application to collate them locally -- after HoT there will be 9 professions with 6 specs each. That's a small enough number that I'm okay with just handing the call to `.reduce` to the application.

> Another small thing we might to consider adding, is how many Hero Points are required to unlock a certain trait. If we keep the ordering of traits the same as the order of unlocking, then we would only have to add the amount of Hero Points needed to provide this data.

I've been fiddling with this and providing a `progression` array with each specialization, e.g., 

```json
{
    "id" : 4,
    "name" : "inventions",
    // ...
    "progression" : [
        { "cost" : 2, "trait" : 432 },
        { "cost" : 3, "trait" : 514 },
        // ...
    ]
}
```

I don't really like it though, since there are progression lines that *aren't* specializations (e.g., the ones you use to unlock skills, which presumably would have `skill` instead of `trait` as the field name in the above example).

It seems really kind of silly to me, but it might make more sense to have yet another endpoint to expose hero point ... trait ... lines? I don't even know the official term for them. Then each specialization can just refer to the accompanying progression thing by id. It's something we should expose, I'm just not convinced that this endpoint is the best place for it.

StrategistXIV:
If you would make a new API (which isn't a bad choice if you ask me!) I would maybe call it /v2/progressionline/ or /v2/heropoints/ or something. I think the last one reads easier but is less clear on what the API is about.

Eearslya:
Will this include revenant at any point before Heart of Thorns?

lye:
Uhh, hmm, not sure why revenant specializations aren't listed there.

Probably not going to be fixed until the next patch though.

PR #43, opened 2015-06-24
Proposal of v2/chat

StrategistXIV:
I bet this is a pretty trick API to implement, but I think it could greatly enhance the GW2 Guild and general community. The API would basically consist of 3 API endpoints

`/v2/chat/receive` - This endpoint receives all the Guild and whisper messages that are cached for a specific amount of time.
`/v2/chat/send/whisper` - This endpoint allows users to send whispers
`/v2/chat/send/guild` - This endpoint allows users to send messages in Guild chats.

There are a few things I am still not sure about:

- Is this implementable at all? (Especially the sending part of the messages)
- Which character is displayed when sending the messages.
- Which formats for strings and time to use. (I can imagine some talk in a language with weird signs, so we should probably at least use UTF-16 if I remember correctly, I am not sure how detailed the time internally is stored)
- For how long are things cached, it would be great if it could be cached for maybe a day, or a few hours.

I wonder what you guys think about this.

### DIFF
--- a/v2/chat/char.js
+++ b/v2/chat/char.js
@@ -1,0 +1,2 @@
+// GET /v2/chat
+["receive","send"]
--- a/v2/chat/receive/receive.js
+++ b/v2/chat/receive/receive.js
@@ -1,0 +1,18 @@
+// GET /v2/chat/receive
+// Authentication API Key required.
+
+[
+	{
+		"Time": <Time format>,
+		"Character": "Charizard",
+		"Content": "Hey, how are you doing guys?",
+		"Type": "Guild",
+		"GUID": <GUID>
+	},
+	{
+		"Time": <Time format>,
+		"Character": "Power Ranger",
+		"Content": "Hey! I am fine, how are you?",
+		"Type": "Whisper",
+	}
+]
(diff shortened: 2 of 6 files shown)


Teranas:
As mention by @darthmaim at PR #25: A chat API is something that was already discussed. (See https://forum-en.guildwars2.com/forum/community/api/API-CDI-2015/page/2#post4813732)

However, i don't think that HTTP(S) is the right protocol to provide a chat api. It's not designed for short messages and instant messaging.

Other protocols like IRC or XMPP fit much better for that use case.

StrategistXIV:
Ah my bad, feel free to shut down this PR if it doesn't stand a chance. I searched at the old PR's but I did not find any in relation to this, so that is why I created a new PR.

darthmaim:
Read https://forum-en.guildwars2.com/forum/community/api/API-CDI-2015/page/2#post4815827 and the next ~10 posts about the protocol discussion.

TL;DR: IRC/XMPP/WebSockets

lye:
Server-sent events is also on the table.

Working chat endpoints are fairly high cost though, since the current API backend is designed around a request/response model rather than keeping open connections. I'd rather leave it until after all of the lower hanging fruit is implemented (and there's still a lot of that to do).

tivac:
Thanks for your contribution!

I completely agree with @lye though, there's so many useful & __way__ easier things we should work on first.

Going to close this for now, not because we won't do it, but because it's far enough out that we should wait until we're able to give it the attention it needs before spending too much time thinking about it.

PR #49, opened 2015-07-01
Add crafting disciplines to /v2/characters

lye:
These would be accessible with the `characters` permission -- I don't think there's much need for an additional permission here.

Before anyone asks, recipe unlocks will be a separate PR.


darthmaim:
Instead of closing this should have been merged?

lye:
Arghghghg waffles.io you have betrayed my trust.

PR #51, opened 2015-07-03
addition to v2/colors

codemasher:
Added tone, material and set as per this forum topic: https://forum-en.guildwars2.com/forum/community/api/API-Suggestion-Colors-API/2910450

Edit: ok, i won't ever manage to handle pull requests... 
Edit2: darthmaim fixed it! ;)

### DIFF
--- a/v2/colors/colors.js
+++ b/v2/colors/colors.js
@@ -1,0 +1,55 @@
+{
+	"id": 126,
+	"name": "Hot Pink",
+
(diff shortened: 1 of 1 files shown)


lye:
Ah, that's a really good idea; I'll have to figure out how those categories are decided (I have a feeling they're derived from the material components -- not manually assigned).

lye:
This has been superseded by #88. Same data, slightly different format. I think the backend parts will go live with the next time we do a patch (maybe?).

PR #53, opened 2015-07-07
Add storymodes and orders to chars

Plitzz:
Is it possible the information selected orders and completed story mode for /v2/characters/characters.js

### DIFF
--- a/v2/account/account.js
+++ b/v2/account/account.js
@@ -5,8 +5,6 @@
 	"id"     : "account-guid",
 	"name"   : "Lawton.1234",
 	"world"  : 1007,
-	"commander": "yes or no",
-	"fractallevel": 11,
 	"guilds" : [
 		"DA9137CD-3A86-E411-B57A-00224D566B58",
 		"1F5F70AA-1DB6-E411-A2C4-00224D566B58",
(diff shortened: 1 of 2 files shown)


lye:
Also needs to list which explorable paths you've done for each dungeon. I don't yet have any immediate thoughts on the response format.

Aralicia:
On crafting : I believe it would be more interesting to have something like : 
    
    crafting : {
        leatherworker : {
            level : 500,
            active : 1
        },
        chef : {
            level : 400,
            active : 1
        },
        tailor : {
            active : 123,
            active : 0
        }
    }

That way, we get both the actual level in the discipline and if the discipline is currently active on this char.

lye:
> On crafting

Check out [this PR](https://github.com/arenanet/api-cdi/pull/49).

Aralicia:
Oups, my bad. I replied because the last comment was a few days old, but didn't see that the "crafting commit" was nearly a month old.
Sorry !

lye:
No biggie; it's tough to keep everything organized.

Plitzz:
Hello there,

It is estimated to when we can use this feature?

Best regards

Plitzz

lye:
> It is estimated to when we can use this feature?

Nope, there is no estimation.

PR #55, opened 2015-07-07
additions to characters.json

codemasher:
A couple of additions to characters.json as per this forum thread: https://forum-en.guildwars2.com/forum/community/api/Character-position/5254020

### DIFF
--- a/v2/characters/characters.js
+++ b/v2/characters/characters.js
@@ -15,7 +15,20 @@
 	guild: "1F5F70AA-1DB6-E411-A2C4-00224D566B58",
 	created: "2015-06-05T19:45:00Z",
 	age: 91, // in seconds.
-	deaths: 0
+	deaths: 0,
+
+	// most recent log-in time
+	"connected": "2015-07-07T05:16:00Z",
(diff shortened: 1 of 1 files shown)


lye:
That pile of data is, uhh, stored in different servers with different TTLs; roughly:

 * Online status (incl. which map and character is logged in) is in the "online status server" and is up-to-date within ~30 seconds.
 * Supply info is owned by the map instance server; the API's read-only snapshots are only updated every 5 minutes.

So for the WvW commander bit, it's totally infeasible. The rest of it is fine, but it needs to be put behind a new scope (e.g., "presence" or something) since I imagine there's quite a few people who don't want to grant access to that.

codemasher:
Would it be possible to add the supply count to the mumble link data instead? This way we'd have in realtime, too.

lye:
I think we've still got a few bytes to spare in there; I'll split that out to a separate PR tomorrow.

lye:
After awhile looking at this, all of these roughly fall under #127 so I'm going to close this one out.

PR #56, opened 2015-07-07
Inventory/Bank: changed upgrades and infusions from array to object

codemasher:
For consistencies sake. This way we can grab all item IDs by recursively searching for the key "id" or for example use functions like [array_column()](http://php.net/manual/function.array-column.php).

### DIFF
--- a/v2/account/bank.js
+++ b/v2/account/bank.js
@@ -8,7 +8,9 @@
 		slot: 1,
 		count: 1,
 		upgrades: [
-			24675
+			{
+				id: 24675
+			}
 		]
 	},
 	{
@@ -24,7 +26,9 @@
 		count: 1,
 		skin: 3706,
 		upgrades: [
-			24661
+			{
+				id: 24661
+			}
 		]
 	},
 	null,
@@ -37,8 +41,13 @@
 		slot: 10,
 		count: 1
 		infusions: [
-			1234,
-			5678
+			{
+				id: 1234
+			},
+			{
+				id: 5678
+			}
+
 		]
 	},
 	{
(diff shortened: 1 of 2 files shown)


lye:
I'm not sure how much I like this; it really depends on whether or not you consider applied upgrades to be "item stacks", which is how most other things are encoded. They can neither stack nor have upgrades/infusions applied to them, which makes me feel that they don't need to use the item stack encoding.

That, and I'm a bit concerned about breaking backwards compatibility.

FrederikNS:
I was working on something using the item api, and this was probably the
one thing that seemed most weird.

Switching to an object is a good idea in my opinion.

darthmaim:
> I'm not sure how much I like this; it really depends on whether or not you consider applied upgrades to be "item stacks", which is how most other things are encoded. They can neither stack nor have upgrades/infusions applied to them, which makes me feel that they don't need to use the item stack encoding.
> That, and I'm a bit concerned about breaking backwards compatibility

Haha, yesterday I started to write exactly the same comment, but somehow forgot to finish and send it.

I wouldn't break backwards compatibility for this, I'm also not seeing how `$upgradeIds = array_column($itemStack->upgrades, 'ids')` is any simpler than just `$upgradeIds = $itemStack->upgrades`.

TL;DR: :-1:

codemasher:
So after a conversation with darthmaim we agreed that just "id" or just an array of numbers isn't the best thing since there are different kind of IDs in the response. This way we're still not sure which kind of ID whe have. Cliff said once that the API should be self explaining - using "item_id", "skin_id" or "whatever_id" would be exactly that.

PR #57, opened 2015-07-09
Change to v2/floors/sectors.json

codemasher:
A feature which has been requested by some wiki editors: sector boundaries as polylines so that we can draw maps like this: http://wiki.guildwars2.com/wiki/File:Queensdale_map.jpg


lye:
Seems reasonable. Good call on the leaflet-compatible format.

PR #60, opened 2015-07-12
Create dungeon.js

dlonie:
Adds dungeon completion info to /account/, and should also be used under /character/. Extends and replaces #53.

The in-game UI provides no mechanism to know which dungeons have been completed for daily rewards, story mode unlocks, or Dungeon Master (DM) completion. Hopefully the API team and the dev community will fill in this missing functionality.

At a minimum, the 'last completion timestamp' is the most important bit, as that will let us know everything we need to know:

- Have I run this path for DM yet?
- Do I need to unlock story mode on this character?
- Have I already run this since the last reset?

### DIFF
--- a/v2/account/dungeon.js
+++ b/v2/account/dungeon.js
@@ -1,0 +1,29 @@
+// GET /v2/account/dungeon
+// Authorization: Bearer oauth2-token
+// Requires "account" scope.
+// This exact format should also have an endpoint under character, for
+// determining, e.g., which story mode unlocks are needed per character.
+{
+  "Ascalonian Catacombs": { // Name of dungeon
+    "Story": { // Path id, either "Story" or "Path X". Alternatively, "Hodgins" "Tzaark", etc.
+      "first": "<timestamp>", // Timestamp of when dungeon was first completed.
+      "last": "<timestamp>", // Timestamp of most recent completion.
+      "count": 4 // Total number of times dungeon has been completed.
+    },
+    "Path 1": {
+      "first": ...
+    },
+    ...
+  },
+  "Twilight Arbor": { ... },
+  ...
+  "Fractals Of The Mist": { // Fractals too!
+    "Scale 1:" {
+      "first": ...
+    },
+    ...,
+    "Scale 50": {
+      ...
+    }
+  }
+}


Plitzz:
I do not think that this is feasible because:

- The story mode for each character is stored permanently
- The exploration mode, however, as far as I know this is true and this is far account only valid until the daily reset
- Saving the way of Verliesmeistersist in the game just is not provided

The API team can, as I understand this, just take data from the game, which are also present in the game, but do not make queries to non-existent data.

dlonie:
I'm having trouble interpreting your response.

- What do you mean by "permanently"?
- What are you talking about being true for explorable modes? I seem to be missing part of the discussion.
- What is Verliesmeistersist? Google / Google Translate are failing me.

Plitzz:
- Each character has to complete only once to unlock the exploration mode the story mode, I meant with permanently

- Any way is for the dungeon master once necessary extent correct, but this is probably not stored timestamp, at least not so that you can read with this first and last.

- Here Google Translate has built me ​​a mistake and I was not paying attention, a lack of space and typed too quickly. The point was:
  The paths for the Dungeon Master to save is not provided in the game

To once again return to your points:
 
Have I run this path for DM yet?
- This information has been around not in the game, so should or can they also not be read via the API

Do I need to unlock story mode on this character?
- This information you can generate yourself with PHP or JavaScript if the story mode would be at the characters.

Have I already run this since the last reset?
- Here I am not sure whether ArenaNet wants this so, since many rewards in the dungeons were reduced.


On the whole, we are here to wait for an official info.
Many of your desired information we do not get so in the game interface.

dlonie:
Thanks for the clarifications.

I'm curious, are you knowledgeable about the format and availability of the data on the ArenaNet servers? It sounds like you're assuming a lot about this. Perhaps it would be best to wait for an employee to comment on these items before dismissing them as infeasible.

- Yes, each character must complete story to unlock explorable mode. This is why I propose adding an endpoint to _both_ account and character scopes. The account scope would be cumulative across characters, and would be useful for determining whether the path is still eligible for daily rewards (which are account based). On the other hand, the character endpoint would show only stats for a particular character, which would provide information about story unlocks.

- I'd rather let someone familiar with the actual GW2 data storage make that call. If the timestamps are unavailable, I can revise the response format to something workable.

- The game is absolutely aware of which paths are needed for Dungeon Master -- otherwise, it would not know when to award the achievement. This information is just not exposed through the player interface, which is a large part of the motivation for this request.

- I'm not sure what reward balancing would have to do with exposing an endpoint that identifies paths still eligible for rewards.

While the information is not displayed in the player interface, it is known to their servers and can be exposed via the API. The lack of an existing interface to this information is why I feel it is a good candidate for a new endpoint.

So yes, we must wait for official info. Speculation at this point will just lead to aimless and unnecessary revisions.

lye:
Haven't looked into the dungeon/achievement stuff yet, but it should be feasible.

Archomeda:
Just to throw it in as reference, #443 will add the daily completions for both dungeons and raids.

lye:
Hmm, in retrospect, I should have re-read this before implementing #443 because this has a better output format.

PR #61, opened 2015-07-12
added v2/floors/events

codemasher:
So i'd like to merge the static data from former /v1/event_details.json into a maps response as per this forum thread: https://forum-en.guildwars2.com/forum/community/api/API-Suggestion-Event-Details/2354712
```
/v2/continents/1/floors/1/regions/4/maps/15
/v2/continents/1/floors/1/regions/4/maps/15/events?ids=all
...
```

Also, this: https://github.com/arenanet/api-cdi/pull/2#issuecomment-75172081

### DIFF
--- a/v2/floors/events-location.js
+++ b/v2/floors/events-location.js
@@ -1,0 +1,27 @@
+[
+	// an events.location object may represent one of the following
+	{
+		"type": "circle",
+		"coords": [12058, 13661],
+		"radius": 250
+	},
+	{
+		"type": "polyline",
+		"start": [12418, 12867],
+		"coords": [
+			[12418, 12867],
+			[12410, 12900],
+			[12410, 12934]
+		]
+	},
+	{
+		"type": "polygon",
+		"coords": [
+			[11358, 13107],
+			[11270, 13233],
+			[11288, 13367],
+			[11424, 13407],
+			[11588, 13477]
+		]
+	}
+]
--- a/v2/floors/events.js
+++ b/v2/floors/events.js
@@ -1,0 +1,13 @@
+{
+	"name": "Defeat the shadow behemoth.",
+	"id": "31CEBA08-E44D-472F-81B0-7143D73797F5",
+	"level": 15,
+	// event type according to http://wiki.guildwars2.com/wiki/Dynamic_event#Event_indicators
+	"event_type": "boss",
+	"flags": [
+		"group_event"
+	],
+	"location": {
+		// location object
+	}
+}
--- a/v2/floors/map.js
+++ b/v2/floors/map.js
@@ -57,5 +57,24 @@
             "id" : 526
         }
     },
+    "events": {
+        "31CEBA08-E44D-472F-81B0-7143D73797F5": {
+            "name": "Defeat the shadow behemoth.",
+            "id": "31CEBA08-E44D-472F-81B0-7143D73797F5",
+            "level": 15,
+            "event_type": "boss",
+            "flags": [
+                "group_event"
+            ],
+            "location": {
+                "type": "circle",
+                    "center": [
+                    12058,
+                    13661
+                ],
+                "radius": 250
+            }
+        }
+    },
     "id" : 26
 }


lye:
Eh, I'm not sure how much I like having the same data accessible via two endpoints (both `/v2/floors/events` and `/v2/events`), but I can see why it'd be really useful to have it in the floors response.

> Also, this: #2 (comment)

That's mostly there as an artifact of using the same code to handle multiple endpoints. Almost every endpoint is basically two functions -- one to load the list of all IDs, and another to load the details for a single ID. Those are passed into a function which handles all the pagination etc. in a uniform manner.

I'd argue that the uniformity makes consumption easier (and it does make implementation easier, though the "frontend" component of the API servers is already the easiest bit to modify).

darthmaim:
> I'd argue that the uniformity makes consumption easier

As someone who just wrote a new API wrapper, yes it does!

codemasher:
>Eh, I'm not sure how much I like having the same data accessible via two endpoints (both
/v2/floors/events and /v2/events`), but I can see why it'd be really useful to have it in the floors response.

The Megaservers made  ```/(v1|v2)/events``` obsolete, to the only thing events related left is ```/v1/event_names``` and ```/v1/event_details``` - both are redundant and the latter contains mostly map related data, so just drop ```/v2/events``` and put the related data under floors (unless you have soething really cool for it).

lye:
That's a fairly good point. For some reason I thought that `/v2/events` provided what `/v1/event_details` did -- which it doesn't.

My original concern about duplicated data is thus moot.

ChieftainAlex:
I still think a v2 for events would be nice. The primary reasons for this would be:

1. The events endpoint as it stands is a single monolithic request and so smaller requests would load quicker - in most cases you only want events for a small number of maps
2.  The v1 events endpoint is currently the only useful endpoint on v1 that isn't replicated in some form on v2.
3. In order to convert from the god awful events coordinate system, you need v1 maps as well. (another monolithic entity)

Rather than implement the maps via "v2/floors" or "v2/continents", personally I'd prefer the v2 events endpoint to take the map id instead, so I would advocate "v2/maps/events".

I've pasted a bit of javascript below that converts the current output of v1/event_details to a format I'd like.

```
// Function to convert local event coordinate system into continent coordinate system
function recalc_coords(continent_rect, map_rect, coords){
    return [
        Math.round(continent_rect[0][0]+(continent_rect[1][0]-continent_rect[0][0])*(coords[0]-map_rect[0][0])/(map_rect[1][0]-map_rect[0][0])),
        Math.round(continent_rect[0][1]+(continent_rect[1][1]-continent_rect[0][1])*(1-(coords[1]-map_rect[0][1])/(map_rect[1][1]-map_rect[0][1])))
    ]
}

// Retrieve the maps API (required for continent_rect and map_rect)
$.getJSON('https://api.guildwars2.com/v1/maps.json').done(function(mapsv1){
    var map_events = {};

    // Retrieve the event_details API
    $.getJSON('https://api.guildwars2.com/v1/event_details.json').done(function(data){
        $.each(data['events'], function(k, v) {
            // Save map id for later usage
            var n = v.map_id; // map number

            // Calculate new location and overwrite old coordinate system
            //    also check if event map exists in maps API
            if (n in mapsv1['maps']) {
                var newcenter = recalc_coords(mapsv1['maps'][n]['continent_rect'], mapsv1['maps'][n]['map_rect'], v.location.center);
                newcenter.push(v.location.center[2]);
                v.location.center = newcenter;
            } else {
                // Labyrinthine Cliffs is not in the maps api, hence forget it.
                return
            }

            // Add GUID
            v.guid = k;

            // Remove keys we didn't want
            delete v.map_id;

            // Check if map exists in object, if not create it.
            if (!(n in map_events)) {
                map_events[n] = { "name": "", "events": [] };
            }

            // Add the name if the map exists in the maps API.
            if (n in mapsv1['maps']) {
                map_events[n]['name'] = mapsv1['maps'][n]['map_name'];
            }
            map_events[n]['events'].push(v);
        });

        // Do something with the new object. Like make it an endpoint, or log it to console
        console.log('https://api.guildwars2.com/v2/maps/events?ids=all')
        console.log(map_events);

        // Example query syntax
        // https://api.guildwars2.com/v2/maps/events?ids=all --- this would return all the events for every map with events.
        //                                           --- equivalent to ---> map_events
        // https://api.guildwars2.com/v2/maps/events?ids=16 --- this would return the events for the given map.
        //                                           --- equivalent to ---> map_events[16]
        // https://api.guildwars2.com/v2/maps/events --- this would return a list of map ids with events.
        //                                           --- equivalent to ---> Object.keys(map_events)
    });
});
```

PR #62, opened 2015-07-12
GeoJSON

codemasher:
http://gw2.chillerlan.net/examples/leaflet-maps-v2-geojson.html

However, i'm a bit unhappy with the map response. Once we get [sector boundarys](https://github.com/arenanet/api-cdi/pull/57) and maybe [event data](https://github.com/arenanet/api-cdi/pull/61), it'd require additional features for each polyline/polygon.

### DIFF
--- a/v2/floors/map-geojson.json
+++ b/v2/floors/map-geojson.json
@@ -1,0 +1,155 @@
+{
+	"type": "FeatureCollection",
+	// [continent_rect[0][0], continent_rect[0][1], continent_rect[1][0], continent_rect[1][1]]
+	"bbox": [9856, 11648, 13440, 14080],
+	"features": [
+		// a map label, may only occur once
+		{
+			"type": "Feature",
+			"geometry": {
+				"type": "Point",
+				"coordinates": [11612, 12888]
+			},
+			"properties": {
+				"id": 15,
+				"type": "label",
+				"name": "Queensdale",
+				"default_floor": 1,
+				"min_level": 1,
+				"max_level": 15
+			}
+		},
+		// landmark
+		{
+			"type": "Feature",
+			"geometry": {
+				"type": "Point",
+				"coordinates": [10062.7, 12022.7]
+			},
+			"properties": {
+				"id": 127,
+				"type": "landmark",
+				"name": "Eda's Orchard"
+			}
+		},
+		// waypoint
+		{
+			"type": "Feature",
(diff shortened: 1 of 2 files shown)


lye:
I might be wrong, but I believe all map labels are actually sector or region names (including, for example, the labels on the WvW maps).

For the `unlock` features -- we should be able to pass the type. We can also pass the render icon URL, if there's a place to jam that in (I'm not familiar with the GeoJSON spec). I *think* that, in addition to dungeons, we might also be able to return vendor/tradingpost/bank/etc icon locations, but I'd have to double-check that that data isn't crazy.

I'm really not a fan of including all languages -- the way the backend servers are set up, each language is an additional request (because it has to switch languages in the same manner as the client) -- which is fairly wasteful if the application is only going to display one of them. Now that we're honoring the `Accept-Language` header, it should almost work automatically for browser-originated requests.

codemasher:
The map/region labels (names) are in the root of the response, while sectors are under "points_of_interest" - so we need a Feature for these in GeoJSON. You can basically put anything into a ```properties``` object. See: http://geojson.org/geojson-spec.html
NPC locations - that'd be awesome! :D

lye:
It's less NPC locations and more "where the markers for NPCs are placed on the worldmap". I *believe* that the marker locations are derived from the NPC locations, but I'm not entirely sure.

codemasher:
Just brainstorming:

I wonder if it's possible to gather all the data which is needed to draw a "clean" world map - without instances etc. y'know, just what we see when we press "M" ingame - and put it to a new endpoint, for example ```/v2/geojson``` so that we could easily pull the data for either a whole worldmap or single regions/maps without digging through ```/v2/continents``` like

```
/continent/1                   //full worldmap for tyria
/continent/1/regions/?ids=1,2
/continent/2/maps/?ids=38,94,95,96
```

The idea goes further: the ```properties``` could hold a bunch more information (if any more available!?).

This would so help the wiki efforts for interactive maps. (leaflet+GeoJSON ftw!)

codemasher:
adding this here for uhmm... reference. https://gist.github.com/codemasher/713e2505bbd517e4530bc59a48b5cb60

codemasher:
a basic google maps implementation
https://github.com/codemasher/gw2map
https://rawgit.com/codemasher/gw2map/master/public/index.html

PR #66, opened 2015-07-24
Get recipe sheet item id from recipe id

RedGlow:
There is now way right now, AFAIK, to get the recipe sheet item id(s) starting from a recipe id, that is, the item which allows me to learn the recipe.

It would be a reverse lookup on the field `details.recipe_id` of a recipe item (e.g.: https://api.guildwars2.com/v2/items/9626), just as the recipe search interface (recipes/search?output=...) is a reverse lookup on the field `output_item_id` of a recipe.


lye:
Yeah, there's a massive need for a generalized query system for both items and recipes (and probably other endpoints later). I spent quite a bit of time looking into this six months ago, and it's really gross to set up with our existing infrastructure. It's still something I want to get done, but it's totally fallen off my schedule at this point.

PR #67, opened 2015-07-27
Initial pass at /v2/traits.

lye:
This is a fairly involved endpoint -- it returns a fairly varied data format. Best to read the markdown [in HTML](https://github.com/arenanet/api-cdi/blob/add-v2-traits/v2/traits/traits.md), I think.

There's a dearth of examples, so at some point this week I'm going to turn on `/v2/traits-beta` which is the current WIP implementation of the endpoint, so that you guys can poke around at it, so we can get some good feedback.

### DIFF
--- a/v2/traits/traits.md
+++ b/v2/traits/traits.md
@@ -1,0 +1,427 @@
+`/v2/traits` acts like a normal bulk-expanded endpoint. When requested with no parameters, it returns a list of ids for every trait object it exposes. These ids can be resolved via a variety of means:
+
+ * `/v2/traits/:id` returns a single trait.
+ * `/v2/traits?ids=2,3,4` returns a set of user-specified traits.
+ * `/v2/traits?page=0&page_size=200` returns a paginated list of traits.
+ 
+Additional metadata is contained in the response headers:
+
+ * `X-Page-Size` (when using `?page=`) echos the requested page size.
+ * `X-Page-Total` (when using `?page=`) contains the total number of pages.
+ * `X-Result-Count` contains the number of objects returned in this response.
+ * `X-Result-Total` contains the total number of objects.
+ * `Link` (when using `?page=`) contains an RFC6690 link specifier.
+
+These should all be exposed to CORS requests.
+
+The following describes the result format.
+
+## Traits
+
+Returned traits will have the following structure --
+
+```
+{
+    id: 238,
+    tier: 3,
+    name: "Evasive Arcana",
+    description: "Create an attunement-based spell after dodging.",
+    slot: "Major",
+    facts: <skill-fact-list>,
+    traited_facts: <skill-fact-list>,
+    skills: <skill-list>
+    specialization: 11,
+    icon: "url"
+}
+```
+
+Where:
+
+ * `id` is the numerical id of the trait.
+ * `tier` is a value from `1-3` and indicates which set of three traits within the specialization the trait can be selected from.
+ * `name` is the localized name of the trait.
+ * `description` is the localized description of the trait.
+ * `slot` is either `"Major"` or `"Minor"`; `"Minor"` traits are the ones that are unselectable.
+ * `facts` are the baseline tooltips that are displayed for the trait.
+ * `traited_facts` are conditional modifications to `facts` when there's interplay between traits.
+ * `skills` is a list of skills that may be triggered by this trait.
+ * `specialization` is a reference to `/v2/specializations` to which this trait belongs.
+ * `icon` is a URL to the in-game icon displayed for this trait.
+ 
+## Skills
+ 
+ The skills returned in `trait.skills` look like --
+ 
+ ```
+{
+    name: "Flame Burst",
+    facts: <skill-fact-list>,
+    description: "Burn nearby foes.",
+    traited_facts: <skill-fact-list>,
+    icon: "url",
+    id: 5794
+}
+```
+
+Where:
+
+ * `name` is the name of the skill.
+ * `facts` are the baseline tooltips that are displayed for the skill.
+ * `traited_facts` are the conditional modifications to facts when the skill is traited.
+ * `icon` is a URL to the in-game icon displayed for this skill.
+ * `id` is the numerical id of the skill.
+ 
+## Traited Facts
+
+The skill facts contained within `trait.traited_facts` and `skill.traited_facts` are the same as normal facts, except they have an additional `requires_trait` field and may have an `overrides` field. `requires_trait` contains a trait id which, when equipped, activates this fact. When activated, if `overrides` is defined it contains the index into the `facts` array that should be replaced by this fact. If `overrides` is undefined, this fact should be appended to the facts list.
+
+As an example (note, chopped up for clarity):
+
+```
+{
+    name: "Aim-Assisted Rocket",
+    facts: [
+        {
+            text: "Range",
+            value: 1500
+        },
+        {
+            text: "Damage",
+            hit_count: 1
+        },
+        {
+            text: "Number of Targets",
+            value: 3
+        },
+        {
+            text: "Radius",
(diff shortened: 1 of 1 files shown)


lye:
So... just turned on [/v2/traits-beta](https://api.guildwars2.com/v2/traits-beta). It's got a bunch of data in it in the format described.

Thoughts?

FrederikNS:
{"text":"API not active"} 😞

lye:
Posted as soon as I committed the config change -- sometimes takes a few minutes to propagate out to the servers :3

Eearslya:
Did you turn on /v2/specializations too so we can reference the specialization ID?

lye:
I'll look into getting the specializations stuff up tomorrow (or alternatively Friday at 5PM) -- the data coming off that is straightforward enough that I don't think it'll change in a large way. Gotta get permission first though.

Eearslya:
Still not active, it seems..How long do these changes usually take?

lye:
Well, this is a real head scratcher. The config update went out half an hour ago, and the API servers are reporting the correct state (that `/v2/traits-beta` should be enabled) but are still kicking back errors about it not being enabled. Grr.

tivac:
@lye figured it out, fix incoming.

lye:
Okay, this should actually be working now.

:shipit:

Eearslya:
So far, after playing with the endpoint for an hour or so, the only thing I find myself wanting is a way to organize them in their tiers; a way to know which trait is top, middle or bottom in the tier, for easy display on webpages.

Eearslya:
After digging into the API a bit more, I've also noticed that some of the traits have traited facts that require..themselves? Specifically, `517`, `604`, and `686` have traited facts that point back at themselves. Is this intentional or some sort of bug with the way traits are coded?

lye:
Hahaha that's just how it is; I can't really make a comment on why.

There's one engineer trait whose skill is affected by the warrior shield
block trait. That one in particular I strip out on the frontend because
it's a bit malformed and breaks my unit tests.

:D

tivac:
Game content is sometimes... unusually structured.

lye:
> Did you turn on /v2/specializations too so we can reference the specialization ID?

Just turned [it](https://api.guildwars2.com/v2/specializations?ids=all) on.

Eearslya:
>I'd have to glance at the game code, but my unchecked supposition is that they're ordered in the order they appear in the specialization `major_traits` list.

It seems you'd be right! However, one quirk I've noticed is with the `minor_traits` list. For 3 of the elite specs (Dragonhunter, Chronomancer, and Reaper) the 4th entry in that list is the 'weapon proficiency' trait. But for Tempest, the weapon trait is first.

I'm basically accessing the array as `minor_traits[tier]` to get each tier's minor; so is this something that should be changed on my end or yours?

lye:
> I'm basically accessing the array as minor_traits[tier] to get each tier's minor; so is this something that should be changed on my end or yours?

Hrm, I think the ordering of `minor_traits` isn't actually meaningful, since there's a 1:1 mapping between tier and traits. The weapon proficiency minor traits all have `.tier = 0` which I think indicates they shouldn't be displayed in the UI? They were displayed at one point, but don't appear to be in the POI videos -- I'm almost thinking they should be removed from the `minor_traits` list.

Eearslya:
>I think the ordering of `minor_traits` isn't actually meaningful (or only coincidentally so), since there's a 1:1 mapping between tier and traits

Well, if `major_traits` is meant to be ordered in the correct format as shown in game, I don't see why `minor_traits` shouldn't be, too. It just makes it a little easier to lay out a webpage format dynamically, without a local database of data.

dr-ishmael:
Trait recharge times are not shown - was this just overlooked? I see "Recharge" type facts in the linked skills, but none at the trait level. Example: [Imbued Melodies](https://api.guildwars2.com/v2/traits-beta/1839) should have a 10 s recharge, according to [the wiki](http://wiki.guildwars2.com/wiki/Imbued_Melodies).

traits.md mentions the ambiguity with "AttributeAdjust" facts in regards to Healing, but what I'm seeing doesn't square with what's stated there.

Examples:
* [Energy Amplifier](https://api.guildwars2.com/v2/traits-beta/519) ([wiki](http://wiki.guildwars2.com/wiki/Energy_Amplifier))  - increases Healing Power attribute
  * {"text": "Attribute Adjust","type": "AttributeAdjust","icon": "...","value": 250,"target": "Healing"}
* [Elemental Bastion](https://api.guildwars2.com/v2/traits-beta/1986) ([wiki](http://wiki.guildwars2.com/wiki/Elemental_Bastion))  - heals
  * {"text": "Healing","type": "AttributeAdjust","icon": "...","value": 552,"target": "Healing"}

Both of them have "target":"Healing" while "text" is different. However, the doc looks like it's saying that "target" should be the one that indicates the difference somehow.

lye:
> Trait recharge times are not shown - was this just overlooked?

...huh, I'll have to look into that.

> However, the doc looks like it's saying that "target" should be the one that indicates the difference somehow.

Yeah, for all types, `text` is an arbitrary unstructured string. The designers put whatever they want there -- it's just for informational purposes. The `target` *should* indicate which attribute it adjust -- but healing power and healing should be distinct. I'll look into fixing that -- it's probably just an oversight on my part.

dr-ishmael:
> (EDIT: looked into the latter issue. For reasons I'm only wont to share over a beer, the healing power vs. health restore attribute adjustment ambiguity isn't something I'll be able to fix in the short term).

I'd happily share a few beers with you, especially since I discovered even more ambiguity within that fact type.

* Life siphon damage is an "AttributeAdjust" fact with "target":"Power" (life siphon heal facts look like standard heal facts).
* Facts where damage is added to non-damaging conditions ([Terror](https://api.guildwars2.com/v2/traits-beta/812) and [Deathly Chill](https://api.guildwars2.com/v2/traits-beta/1919)) show that damage as "AttributeAdjust" facts with "target":"ConditionDamage".

lye:
Yeah, those are totally ambiguous in the source data as well.

In all fairness, this data wasn't really meant to be read mechanically -- it's to display the trait/skill tooltips -- so it kind of makes sense that it's a bit of a mess :frowning:

Lootballs:
When damage is used in traits would it be possible to get the coefficient used to calculate it (the same for healing)? Like a 0.2 scaling for example. I looked through a few and couldn't see it.

lye:
Damage coefficients are currently stripped from the output -- I meant to figure out whom to ask permission to display 'em but it slipped my mind. I'll renew my quest on Monday and see if whether or not we can display those.

Lootballs:
They are already on the wiki, adding a layer of obscurity doesn't help anybody in my opinion.

dr-ishmael:
They're only on the wiki because players have researched them. While I
don't do that myself, I'm sure that for those who do, it gets annoying
having to do it all again after any update that adjusts a skill's damage.
Also, it can be confusing when someone doesn't understand the conditions
necessary to get a "clean" coefficient (e.g. can't have anything that
increases your Power above the base).

I don't see any reason why these coefficients couldn't be shown in the API,
and it would be a great help to the players to have a fully consistent
source for them.

lye:
So skill damage coefficients and the somewhat unavoidable and unfortunate ambiguities aside, it doesn't seem like there are too many other outstanding issues with `/v2/traits-beta`. If that's the case, I think it makes sense to turn on `/v2/traits` on Monday, Aug 24 and probably turn off `/v2/traits-beta` on Friday, Aug 28.

I still need to bounce the coefficient issue around a bit, but since that's an additive change I don't think it's a blocking concern.

darthmaim:
Now that /v2/traits is live this should be merged.

lye:
> 21 days ago

I am good at this, team.

PR #79, opened 2015-08-09
addition of currency to bank.json

bobbyjoemoma:
Accessing the Bank endpoint will provide the amount of deposited money in the bank as well as the previously returned array of items in the banks storage.

Should be easy to get this information? But I also might not know what I am talking about :)

### DIFF
--- a/v2/account/bank.js
+++ b/v2/account/bank.js
@@ -1,71 +1,75 @@
 // GET /v2/account/bank
 // Authorization: Bearer oauth2-token
 // Requires "account" and "inventories" scopes.
-[
-	null,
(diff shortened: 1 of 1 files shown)


RedGlow:
It doesn't look really clean to me.

It mixes kinds of information which are distinct at a conceptual level (bank content and currencies). I think that https://github.com/arenanet/api-cdi/pull/80 is a better way to expose the currencies, and the rest of the bank access is good the way it already is.

bobbyjoemoma:
ya I pushed it and then I thought wait this is not very good at all... haha

PR #80, opened 2015-08-10
create account/wallet.json

milneti:
This endpoint would be used to gather the current status of the players wallet. wallet is account wide so they should only need account/inventory scopes. 

I am not sure the form of how currency is stored... i.e. is there a name/type/value associated with them or is this stored another way? I assumed it was some form of this... I can modify the outline as necessary if this information is not available in this form.

I believe this data should be easily accessed? Or I may be completely wrong :)

### DIFF
--- a/v2/account/wallet.js
+++ b/v2/account/wallet.js
@@ -1,0 +1,49 @@
+// GET /v2/account/wallet
+// Authorization: Bearer oauth2-token
+// Requires "account" and "inventories" scopes.
+[
+	{
+        //name of the currency
+        name: "gems",
+        
+        //type of currency
+        type: "general",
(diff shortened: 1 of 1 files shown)


lye:
The game doesn't actually make a distinction between types of currencies -- in fact, most currencies are a map currency in that they appear on the bottom-left of the inventory UI on at least one map.

My plan for the wallet was to expose two endpoints -- `/v2/currencies` which provides the id, localized name and description of all the currencies and `/v2/account/wallet` which provides a list of `(id, quantity)` locale-independent tuples. That way you can display it in whatever language.

lye:
See #84.

PR #84, opened 2015-08-21
Add in /v2/account/wallet and /v2/currencies.

lye:
These are pretty straightforward endpoints that expose the list of all
currencies that can appear in the player's inventory, and an authenticated
endpoint to show the amount an account has of each.

Supercedes #80.


KelianB:
When can we expect a release of these endpoints? 
Which ones will come first, wallet & currencies or dyes & skins?

lye:
The backend bits are tied to the game release process, so whenever they get shipped to live. Will probably release them all at once, provided there's no issues.

PR #75, opened 2015-08-03
Add specializations to /v2/characters

Eearslya:
Noticed there wasn't already a PR for this, so I went and made my own.

Nice and simple, exposing each character's currently equipped specs.

PR #85, opened 2015-08-22
Add specializations to /v2/characters.

lye:
This is a slight change to #75; it's exactly the same except it provides
access to builds from all three game modes (since they're stored
differently now).

### DIFF
--- a/v2/characters/characters.js
+++ b/v2/characters/characters.js
@@ -134,6 +134,130 @@
 	]
 }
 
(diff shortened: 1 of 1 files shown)


madou:
As a side.

What do we think about getting equipped pvp gear?

E.g. this data: https://d3b4yo2b5lbfy.cloudfront.net/wp-content/uploads/2014/03/42918build-ui.jpg

PR #88, opened 2015-09-01
Add item/categories to /v2/colors.

lye:
* `item` contains the item id which unlocks the color.
 * `categories` contains the in-game sort categories.

[Forum thread](https://forum-en.guildwars2.com/forum/community/api/v2-colors-Question).

### DIFF
--- a/v2/colors.js
+++ b/v2/colors.js
@@ -49,8 +49,19 @@
 			193,
 			192
 		]
-	}
+	},
+	"item": 20359,
+	"categories": [
+		"Gray",
+		"Metal",
+		"Rare"
+	]
 }
+
+// Possible categories:
+// Hue: ["Gray", "Brown", "Red", "Orange", "Yellow", "Green", "Blue", "Purple"]
+// Material: ["Vibrant", "Leather", "Metal"]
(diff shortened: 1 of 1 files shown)


tivac:
:+1: I was just wishing for this info a couple of nights ago.

StevenLiekens:
I don't like how "categories" is a single field that represents three different concepts (hue, material, rarity). Using an array is weird because a color can only have 0..1 of each category. 

IMO this should have looked different.
```js
{
	"id": 6,
	"name": "Celestial",
	"base_rgb": [
		128,
		26,
		26
	],
	"cloth": {},
	"leather": {},
	"metal": {},
	"item": 20359,
	"hue": "Gray",
	"material": "Metal",
	"rarity": "Rare"
}
```

codemasher:
That was already the case in my original PR, also don't like that it's a single fireld. https://github.com/arenanet/api-cdi/pull/51

lye:
Having it as an array is more future-proof: a color may match be in multiple slots of the same category.

Also it's a bit late to change :<

Also I forgot to merge this.

PR #89, opened 2015-09-03
Add /v2/pvp/games.

lye:
This is an initial subset of the features listed in #38.

As an additional note, match retention is very **very** limited. Only up to the **ten** most recent games are stored -- and I think these are persisted for at most a month.

### DIFF
--- a/v2/pvp/games.js
+++ b/v2/pvp/games.js
@@ -1,0 +1,52 @@
+// Bulk-expanded endpoint with GUID keys.
+// Authenticated; requires "pvp" permission.
+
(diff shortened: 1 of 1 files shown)


StrategistXIV:
Looks good to me! Glad to see this initial version. It would be nice if we somehow could add the other players and their build later. But I think its important to first complete the skill API for that.

Good job guys!

lye:
> It would be nice if we somehow could add the other players and their build later.

Rest assured, though it's going to be a lot of work to plumb through all the data, this is something I really really want. All in due time.

PR #90, opened 2015-09-03
Add whitelist status to /v2/commerce/prices

tivac:
So you can check out which items a P4F account can buy/sell.

### DIFF
--- a/v2/commerce/prices.js
+++ b/v2/commerce/prices.js
@@ -1,16 +1,20 @@
 // Normal bulk-expanded endpoint that exposes aggregated price information
-// about items on the Trading Post.
+// about items on the Trading Post. Also shows whether or not they are whitelisted
+// to enable Play For Free accounts to buy or sell them.
 
 // GET /v2/commerce/prices
 [ 1, 2, 3, 4, ... ]
 
 // GET /v2/commerce/prices/5
 {
-    id   : 5,
+    id          : 5,
+    whitelisted : true,
+    
     buys : {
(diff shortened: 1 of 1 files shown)


darthmaim:
Not sure how this is managed in your backend, but wouldn't it make more sense to (also) add this to the /v2/items endpoint (maybe as flag)?

tivac:
I don't think TP whitelist status should live on `/v2/items`, it's a TP-specific concept.

I'd agree with an argument about putting the data on `/v2/commerce/listings` as well but it's more work since `/v2/commerce/prices` already had the whitelist status available to it.

darthmaim:
Sounds reasonable. My background was that I am only querying /v2/items for https://gw2treasures.com/ and currently not using the /v2/commerce endpoints at all, but the whitelist status would maybe be something useful to show for each item.
I probably just need to get around to show the current TP prices as well and get the whitelist status with that, just didn't have time for that yet because I need to rewrite a bunch to make this work (updated way more frequent than item details).

lye:
Hmm, actually now that I think of it, `/v2/items` isn't a terrible place for it either. A bit of redundancy isn't bad (and `/v2/commerce/prices` is a fine place for it, since it's a very lightweight endpoint), but `/v2/items` makes a lot of sense, now that you mention it.

StevenLiekens:
Add it to `/v2/commere/listings` too please.

lye:
This is apparently live I guess?

PR #91, opened 2015-09-10
Add /v2/wvw/objectives.

lye:
This is what I'm working with currently -- it's a step up from `/v1/wvw/objectives.json` which has been totally broken for a long while. I'm a bit on the fence about the `map_type` values -- might just change them to be `RedHome` etc to match what's being returned by `/v1/wvw/match_details.json`.

Thoughts?

### DIFF
--- a/v2/wvw/objectives.js
+++ b/v2/wvw/objectives.js
@@ -8,7 +8,7 @@
 {
 	id       : "95-43",
 	type     : "Camp",
-	map_type : "GreenBorderlands",
+	map_type : "GreenHome",
 	map_id   : 95,
 	name     : "Hero's Lodge",
 	coord    : [
@@ -34,7 +34,7 @@
 	{
 		id       : "94-32",
 		type     : "Keep",
-		map_type : "RedBorderlands",
+		map_type : "RedHome",
 		map_id   : 94,
 		name     : "Etheron Hills",
 		coord    : [
@@ -45,7 +45,7 @@
 	{
 		id       : "38-11",
 		type     : "Tower",
-		map_type : "EternalBattlegrounds",
+		map_type : "Center",
 		map_id   : 38,
 		name     : "Aldon's Ledge",
 		coord    : [
@@ -69,10 +69,10 @@
 //   * Spawn
 //
 // map_type is one of the following:
-//   * EternalBattlegrounds
-//   * BlueBorderlands
-//   * RedBorderlands
-//   * GreenBorderlands
+//   * Center (EB)
+//   * BlueHome (Blue BL)
+//   * RedHome (Red BL) 
+//   * GreenHome (Green BL)
 //   * ObsidianSanctum
 //   * EdgeOfTheMists
 //
--- a/v2/wvw/objectives.js
+++ b/v2/wvw/objectives.js
@@ -1,0 +1,88 @@
+// Bulk-expanded endpoint that provides some data for WvW
+// objectives.
+
+// GET /v2/wvw/objectives
+[ "95-43", "38-17", ... ]
+
+// GET /v2/wvw/objectives?id=95-43
+{
+	id       : "95-43",
+	type     : "Camp",
+	map_type : "GreenBorderlands",
+	map_id   : 95,
+	name     : "Hero's Lodge",
+	coord    : [
+		6931.58,
+		14502.1
+	]
+}
+
+// GET /v2/wvw/objectives?page=0&page_size=3
+[
+	{
+		id : "968-85",
+		type : "Resource",
+		map_type : "EdgeOfTheMists",
+		map_id   : 968,
+		name     : "Tytone Perch Reactor",
+		coord    : [
+			7852.89,
+			9855.56
+		],
+		marker   : "https://render.guildwars2.com/..."
+	},
+	{
+		id       : "94-32",
+		type     : "Keep",
+		map_type : "RedBorderlands",
+		map_id   : 94,
+		name     : "Etheron Hills",
+		coord    : [
+			11212.6,
+			10929.3
+		]
+	},
+	{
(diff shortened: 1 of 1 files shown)


codemasher:
I actually don't care on the value of ```map_type``` as long as i'm able to distinguish the different maps through it ;) However, the *Border and Center are already established, so that would be kind of backward compatible or at least the least confusing.

for the rest: ```git push -f```

bherbst:
I like it!

A few questions- is there a general pattern as to when the marker is available and when it isn't? For example, is it just "resource" objectives, or is it somewhat random?

Similarly, when might an objective not have coordinates?

As for map type- I agree that having it match /match_details would be nice, though I don't have strong feelings either way as long as there is a clear mapping.

codemasher:
As per https://forum-en.guildwars2.com/forum/community/api/Knowing-which-borderland-map-has-a-given-id/5477422

Just drop ```map_type``` then, unless it provides some extra value in any way.

bherbst:
Assuming that the desert borderlands have different map IDs, I would rather not have to maintain an id-color map client-side. It's probably fairly safe to assume that the IDs won't change often/ever, but if they add a map in the future it would be great to not have to update a hard coded mapping.

codemasher:
Idea:

What about an additional parameter ```match_id``` (based on what we have now in /v1, may change) which adds additional info to each objective, like last captured, owner color, guild id/name/tag for the given match.

bherbst:
@lye I came up with some more questions- will this endpoint return *all* the objectives (i.e. from both the desert borderlands and the alpine borderlands), or just those form the currently "active" borderlands?

This also would likely be a good endpoint for caching, since it isn't likely to change often. Have you considered adding cache-control headers or something similar for this endpoint? It might be a nice feature for a future release.

@codemasher I would assume that information would be available in `/wvw/match_details`, which I also assume is coming to v2. To me, this endpoint is more or less static information that won't be changing very often.

codemasher:
> @codemasher I would assume that information would be available in /wvw/match_details, which I also assume is coming to v2. To me, this endpoint is more or less static information that won't be changing very often.

Thats what i'm pointing at. The objectives part of the ```/v1/wvw/match_details.json``` felt always a bit out-of-place despite the fact it was kinda required in case you wanted to calculate the detailed scores. If ```/v2/wvw/match_details``` would just return the detailed scores (see https://github.com/codemasher/api-cdi/commit/2e4c46ed18824d13f5dff21c2fb05f061cb70db8) which could be used to create scoreboards or detailed stats while ```/v2/wvw/objectives``` would return anything nessecary to render a map view.

lye:
> However, the `*Home` and `Center` are already established, so that would be kind of backward compatible or at least the least confusing.

Yeah, I'll update those accordingly to match the old values. Not sure why I used different values, derp. I'm not entirely sure it's useful to have those values in the `/v2/wvw/objectives` endpoint (it might make more sense in `/v2/wvw/maps` or even in `/v2/wvw/matches` by having an `{eb: 94, red: 96, blue: 97, green: 98}` `map_id` mapping) but that's the current plan. Some duplication isn't the end of the world.

> What about an additional parameter `match_id`

The intent of this endpoint is to give match-agnostic data; that data will go into `/v2/wvw/matches` I think.

> The objectives part of the /v1/wvw/match_details.json felt always a bit out-of-place

I still need to do a bunch of work on new backend features for the matches endpoint -- I don't have any qualms breaking up the endpoint to suit a wider variety of use-cases. Let's hold off on that for another PR though.

> A few questions- is there a general pattern as to when the marker is available and when it isn't?

Whenever a designer (or myself) decides to populate the field in the content. The way wvw maps are structured has mutated a bit for the new desert borderlands -- the way it used to work was that the maps were literal copy-pastes and then some fields were changed. The new way is to support a programmatic copy-paste wherein all borderlands have different `map_id`s but all point to the same actual map (which is retrofitted with `map_id -> parameter` mappings for e.g. sector names). Thus since there's only one actual map all the objectives have the same ID across different color borderlands, but different names depending on which `map_id` is used.

I don't remember why that tangent was relevant, but w/e.

Oh. Anyway, the objectives don't have coordinates (or names even!) but are tied to a map sector which has a name/coordinate. I'm not really sure why some sectors don't have names/coords. It's something I'll probably look into at some point, but ATM fairly high-cost since I'm not familiar with how to manipulate the map details on a low-level.

> will this endpoint return all the objectives

Yeah. Including those from EotM, in case you want to use 'em (the special markers are included for the most part). When it initially goes live it won't have the desert borderlands objectives since those aren't in the dat file yet, but it'll automatically update once that content gets pushed out.

bherbst:
Awesome, that all sounds perfect!

:thumbsup:

bherbst:
Oh- is there a convenient way to get the sector associated with an objective to get coordinates? And are those coordinates relatively accurate?

codemasher:
> The intent of this endpoint is to give match-agnostic data; that data will go into ```/v2/wvw/matches``` I think.

It was just an idea to seperate these two. So one could for example synchronize the scoreboard with the score tick while the map gets it's updates say every 30 seconds.

> Oh- is there a convenient way to get the sector associated with an objective to get coordinates?

Once https://github.com/arenanet/api-cdi/pull/57 is live, you could simply use a point-in-polygon implementation.

bherbst:
> Once #57 is live, you could simply use a point-in-polygon implementation.

Assuming I know which sector maps to which objective. Which I could figure out once and be done with it, but I'd rather have a reliable way to map them.

lye:
> Oh- is there a convenient way to get the sector associated with an objective to get coordinates? And are those coordinates relatively accurate?

I should put the sector_ids in there I guess. The coordinates returned are the sector centroids; at some point we'll expose the actual polygons.

PR #94, opened 2015-09-18
Add /v2/wvw/matches

lye:
This is what I've got for the new matches endpoint. This PR supplants #8.

I kind of want to make a `/v2/wvw/overview` or something that provides a `world_id -> match_id` mapping, but I'll save that for a separate PR. I'm also not entirely happy with putting the world MMR in this endpoint (feels like it belongs elsewhere maybe?) but eh.

Anyway, I *think* I've got most of the requested data in the new endpoint (I even got PPK data being piped through -- though Sentries aren't happening).

### DIFF
--- a/v2/wvw/matches.json
+++ b/v2/wvw/matches.json
@@ -1,0 +1,84 @@
+// Bulk-expanded endpoint that provides data for the
+// current WvW state.
+
+// GET /v2/wvw/matches
+[ "1-1", "1-2", "1-3" ]
+
+// GET /v2/wvw/matches/1-1
+// GET /v2/wvw/matches?id=1-1
+// GET /v2/wvw/matches?page=0&page_size=1
+
+{
+	"id"         : "1-1",
+	"start_time" : "2015-09-16T21:00:00Z",
+	"end_time"   : "2015-09-17T21:00:00Z",
+	"scores"     : {
+		"red"   : 44820,
+		"blue"  : 44820,
+		"green" : 44820
+	},
+	"ratings"    : {
+		"red"   : 1497.154175,
+		"blue"  : 1502.117432,
+		"green" : 1498.113525
+	},
+	"worlds"     : {
+		"red"   : 1001,
+		"blue"  : 1002,
+		"green" : 1003
+	},
+	"maps" : [
+		{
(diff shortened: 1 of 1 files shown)


bherbst:
A thousand :+1:  for you!

This looks great! A couple of questions-

1. Are `kills` and `deaths` just the total number of kills each world has made and how many types a player in a given world has died, respectively? If so, that's some cool info I didn't expect ;)

1. I'm assuming shrines will come through as objectives when the desert borderlands is live- is that correct? If so, will anything come through in `bonuses` for the desert borderlands?

1. Will information for the Oasis event just come from the events endpoint?

codemasher:
> I'm also not entirely happy with putting the world MMR in this endpoint (feels like it belongs elsewhere maybe?) but eh.

```/v2/worlds```.

Looks nice! 
One question: will we get detailed borderland scores somewhere or do we need to calculate these still?

lye:
> Are kills and deaths just the total number of kills each world has made and how many types a player in a given world has died, respectively?

`kills` are the number of deaths a member of the server has been responsible for, e.g. if one or more red players gets credit for the death of a blue player, `kills.red += 1` and `deaths.blue += 1`. If someone just dies to an NPC and no enemy players get credit, only `deaths` is incremented.

> I'm assuming shrines will come through as objectives when the desert borderlands is live- is that correct?

I'd like to, but the data isn't currently plumbed through -- it's something I'll be looking into next week. I also want to have Outnumbered in there too, but it's not currently  hooked up.

> . `/v2/worlds`

That's actually a pretty good place for MMR, now that you mention it.

> One question: will we get detailed borderland scores somewhere or do we need to calculate these still?

Both the match and individual map scores are included in the response.

codemasher:
> Both the match and individual map scores are included in the response.

I noticed that, but i meant more like https://github.com/codemasher/api-cdi/commit/2e4c46ed18824d13f5dff21c2fb05f061cb70db8#diff-95b77aff4547911b4d848d62f0053cdeR148

lye:
Ah, hmm, that shouldn't be too hard to add in.

lye:
This was secretly turned online, so I guess I should merge it.

PR #95, opened 2015-09-18
Add chat_link field to /v2/items

tivac:
Would also like to add to `/v2/continents`, `/v2/skills`, `/v2/recipes`, and `/v2/traits` but those don't have example files yet.

### DIFF
--- a/v2/items.js
+++ b/v2/items.js
@@ -23,6 +23,7 @@
     "flags": [ ],
     "restrictions": [ ],
     "id": 12138,
+    "chat_link": "[&AgFqLwAA]",
     "icon": "https://render.guildwars2.com/file/CA45CEE2BA3BA040E7C294965CA3756FC3F33FD1/63248.png"
 
 }


codemasher:
Oh, pregenerated chatcodes for the lazy - isn't that a bit over the top? :D
It'd be actually cooler if the game would fully support itemstack chatcodes (infusions are missing).

tivac:
Chat codes are weird, and the less bit-fiddling people need to understand to get access to them the better.

Barking up the wrong tree for changes to game support though, sorry.

codemasher:
>  the less bit-fiddling people

There are 10 kind of people who understand binary!

(Anyway, fair enough, doesn't hurt :octocat: )

lye:
Also some other things appear to have acquired chat links.

>_>

StevenLiekens:
Does a separate endpoint for chat links sound like a good idea?

`/v2/chatlinks/:type/:id?params`

Examples

 - `/v2/chatlinks/items/72`
 - `/v2/chatlinks/items/72?suffix_item=24687`
 - `/v2/chatlinks/items/12138?quantity=250`
 - `/v2/chatlinks/traits/223`
 - `/v2/chatlinks/coins/100`

Artanis:
@StevenLiekens Also, this is more of a [remote procedure call (RPC)](https://en.wikipedia.org/wiki/Remote_procedure_call): you're giving the server item parameters and it does the calculating for you. There's no way this can be pre-calculated like chatlinks in the item definition; there's too many (4 billion in the coins example alone).

There's no need to have the server do that for us. If you need detailed chatlinks like this, use or write a client-side library for that purpose.

codemasher:
A reference on chatlink/itemstack en-/decoding would be welcome. Implementing on the other hand isn't that hard... bit-fiddling though.
https://github.com/codemasher/gw2-database/blob/master/classes/gw2api.class.php#L264

darthmaim:
Yep, an official reference implementation of chatlinks parsing/creating like the actual game does would be great.

PR #96, opened 2015-09-21
Add /v2/achievements and /v2/achievements/daily.

lye:
This is a first-pass -- for the initial release `/v2/achievements` will only be populated with the current daily achievements. I need to go through the entire list and make sure there's nothing that shouldn't be exposed before we can enable it for all the data.

We'll be extending the achievement endpoint a bit more when we release `/v2/account/achievements` so that for achievements that use a bitset internally for progress (e.g., Dungeon Master) you can reference what the individual bits mean (since `/v2/account/achievements` will just give you the completed bit indexes). That's kind of a longer-term change since it requires changes to the actual underlying content -- we have new support to display strings for the individual achievement steps in-game now.

Anyway, let me know if anything looks terribly off with the structure.

### DIFF
--- a/v2/achievements/daily.js
+++ b/v2/achievements/daily.js
@@ -1,0 +1,34 @@
+// Provides today's daily achievements.
+
+// GET /v2/achievements/daily
+
+{
+	pve : [
+		{
+			id    : 34,
+			level : {
+				min : 1,
+				max : 80
+			}
+		},
+		{
+			id    : 35,
+			level : {
+				min : 11,
+				max : 79
+			}
+		}
+	],
+	pvp : [
+		...
+	],
+	wvw : [
+		...
+	]
+}
+
+// "id" corresponds to the achievement id and can be resolved
+// against /v2/achievements.
+
+// "level" is the min/max character level that this achievement
+// will be attainable for. The range is includive on both ends.
--- a/v2/achievements/daily.js
+++ b/v2/achievements/daily.js
@@ -2,6 +2,10 @@
 
 // GET /v2/achievements/daily
 
+[ "na", "eu" ]
+
+// GET /v2/achievements/daily/{na,eu}
+
 {
 	pve : [
 		{
--- a/v2/achievements/index.js
+++ b/v2/achievements/index.js
@@ -1,0 +1,47 @@
+// Normal bulk-expanded endpoint for fetching achievements.
+
+// GET /v2/achievements
+
+[ 1, 2, 3 ]
+
+// GET /v2/achievements/1
+
+{
+	"id"          : 1,
+	"icon"        : "https://render.guildwars2.com/...",
+	"name"        : "Some Achievement",
+	"description" : "A short text string about this achievement",
+	"requirement" : "What's displayed in-game about the requirements for this",
+	"type"        : "Default",
+	"flags"       : [
+		"Pvp"
+	]
+}
+
+// GET /v2/achievements?ids=1
+// GET /v2/achievements?page=0&page_size=1
+
+[
+	{
+
+	    "id"          : 1859,
+	    "icon"        : "https://render.guildwars2.com/...",
+	    "name"        : "Daily WvW Caravan Disruptor",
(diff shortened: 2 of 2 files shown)


Plitzz:
It runs but for PvE there r no Icon.
look here
http://test.teufelspack.de/achievements

Coffee4cr:
#99 as I posted here, they don't seem to show the proper dailies either, when you compare them to
https://wiki.guildwars2.com/wiki/Daily/Table which also seems 1 day off.

Plitzz:
Ingame checked, does not provide the current, at least not in Europe

Coffee4cr:
Yeah I don't think it's different on NA/EU

I know last night when i did my dailies, that Ruins was one of them, and ascalon lumberer

Plitzz:
Maybe not different, but the time lag could be a problem

lye:
I'm guessing there's time lag between NA/EU, since their daily resets are at different times. Totally forgot about that; unfortunately won't be able to fix that until some time after HoT ;_;

lye:
Updated the PR with my proposed fix for the NA/EU split. Note that this is a breaking change (but I can't figure out a cleaner way to do it).

kasoki:
Any ETA on when the breaking change comes live?

lye:
Not for another couple of weeks -- putting out a couple dozen small fires right now and the release window for making changes to that component is looking kind of gross. Sorry for the delay :(

One change that might be reasonable is to have `/v2/achievements/daily` continue returning the bad data for awhile (rather than `["na", "eu"]`) and add in `/v2/achievements/daily/{na,eu}`. I'm not sure how useful that is, but it's an option so that you can write working applications now, then have a window (2-3+ weeks) to update them without breakage.

kasoki:
That would be really useful ;).

RedGlow:
I'm wondering about the /v2/account/achievements. It's clear you already have it planned, so I won't make a pull request, but do we know when this feature may land? Given the increasing amount of achievement-based collections, it's getting more and more useful.

lye:
Hrm, actually, lemmie see if I can turn it on this week.

themacguffinman:
Are fractal dailies going to be in /v2/achievements/daily?

Also, is the final endpoint going to include reward details?

lye:
Uhh, those are really good points. I should add both of those.

lye:
The `/v2/achievements` is now no longer dailies-only, and `/v2/account/achievements` is online. The backend support for the NA/EU fixes aren't going to land until the Dec. 1st patch -- and hopefully I'll have the rest of the missing achievement details in there -- so I'll open a separate pull request for those.

Eearslya:
Trying to document it on the wiki, noticed a couple things.
Could we potentially get some information in `/v2/achievements` as to exactly what the bitmask means?
Also noticed a flag you hadn't mentioned in your PR, `RepairOnLogin`. (ID `2314`) Is that something we're supposed to see?

lye:
Yeah, the `bits` will make more sense when the changes to `/v2/achievements` goes out -- they basically correspond to indexes into a list of strings indicating which steps for the achievement there are. I left 'em in for the initial release because they're indexes into a string array that you can see in-game.

Uhh, I'll probably remove `RepairOnLogin` later this week. It's not really useful for application developers; left it in by accident.

Eearslya:
Is `bits` only a 0-7 thing? If so, how will that work with achievements that have more than 8?

And are the achievement categories part of the missing information planned to be added?

lye:
The values in `bits` are all `uint32_t`s, so they can go up quite far. They're stored internally as a bit array, hense "bits".

Plitzz:
Hello there,

I regret to inform you that the API provides erroneous data.

I currently only randomly Armorer tested, but here are the most error:

Example Greatsword:
{"id": 57, "current": 1, "max": 5000, "Done": false}

I have the success, however, for a long time.

If I find the time today, I'll make a quick side and then the evening to have today a comparison with the information in the game.

lye:
Yep, found the issue for that. Will try to get a fix out with the next patch ;____;

PR #101, opened 2015-10-09
/v2/wvw/objectives: fix coordinates, add example.

lye:
The coordinates we're currently returning are the label coordinates. These are being moved to the `label_coord` field and the *correct* coordinates for the objective marker are going to the `coord` field. Also included is a quick and dirty sample with leaflet.js that puts the objectives on a map (from the tile service).


lye:
This is live (woot).

PR #107, opened 2015-10-28
Add endpoints to retrieve mini data and unlocks.

lye:
This one's pretty straightforward. There's currently not much data in `/v2/minis`, but I don't want to just jam it into `/v2/items.details` (because there might be more later, and mini ids are probably referenced from other places (e.g., achievements)).

Would like to get this deployed later this week or early next week, so if you see any issues let me know soonish <3

### DIFF
--- a/v2/account/minis.js
+++ b/v2/account/minis.js
@@ -1,0 +1,8 @@
+// Returns a list of mini ids that the account has unlocked; these
+// can be resolved against /v2/minis.
+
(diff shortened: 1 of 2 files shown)


hackedd:
`/v2/minis` seems to be missing from `/v2`...

lye:
WHY DO I ALWAYS DO THIS?

I'll have that fixed later this week.

PR #108, opened 2015-10-28
Add binding information to item stacks.

lye:
(the ones returned from /v2/characters and /v2/account/bank).

Pretty straightforward change, I think. Backend support's going to take awhile to go through the release process though, so there's about a month for suggestions/improvements.


darthmaim:
This is now live and can be merged.

PR #113, opened 2015-11-10
Recipes

RedGlow:
A proposal to get the list of known recipes for each character.

This is useful for various reasons:

- crafting helpers know which recipes need to be bought or not

- find which character has learned a recipe without logging in and out of various crafters


lye:
This kinda fell through the cracks at some point -- there was support in the backend for this, it just wasn't turned on. The output format is slightly different, so I've amended your PR a bit and merged what's on live right now. Sorry for the delay!

PR #103, opened 2015-10-13
Add modules to /v2/account

ChalyFlavour:
Extended information about the game content available for this account. Possible flag should be at least:
gw2_base (regular GuildWars2 game, e.g. bought at guildwars2.com)
gw2_f2p (GuildWars2 f2p edition)
gw2_hot (activated GuildWars2 Heart Of Thorns addon)

Some extended flags may apply, e.g.
gw2_wvw (requirements met for playing on world-vs-world maps)
gw2_guildbank (access to guildbanks are granted)

PR #114, opened 2015-11-11
Add game access level to /v2/account

lye:
This pulls in #103 and adjusts it a bit.

Instead of putting the game access levels into a flags array, just expose the latest version the account has purchased. Each access level implies all previous access levels (e.g., if you've got HeartOfThorns you've also got the base GuildWars2); stuffing it into a single field makes it a bit easier to think about.

These are available with just the `account` permission.


darthmaim:
This is now live and can be merged.

PR #115, opened 2015-11-15
Update Characters Endpoint

milneti:
+include map completion percentage
+include mastery level in characters endpoint
+specify if character has elite specialization unlocked in simple boolean

Map completion is a useful statistic for applications to be able to know for displaying character progression.

Mastery level will be useful to accurately show a character's level and progression as it appears in game.

Being able to identify a character having their elite specialization unlocked will allow for showing correct profession icons and allow applications to know the characters status without having to parse through all of their specializations and determine if an elite one exists.

If there is a better suggestion as to the inclusion of these items, please feel free to suggest.

Thanks!


v2/characters/characters.js
```diff
@@ -11,7 +11,10 @@
 	race: "Human",
 	gender: "Female",
 	profession: "Thief",
+	elite: true,
```
Fishrock123: Should this leave room for future options?

madou:
Considering when you select an elite spec you're essentially changing class, why not just allow the profession to be the name of the elite spec if selected?

dr-ishmael:
Anet does plan to release additional elite specializations in the future,
so making the elite status a boolean would not be futureproof.

milneti:
Definitely understand the lack of being future proof. I guess that leaves some options:

If only one elite spec can be active at once, then having the profession return the active spec (or the profession if no spec is active) is a great way to do so. However this could be breaking for other applications. Perhaps still having the "elite" field but replace the boolean with the spec name would be a safer deploy and still accomplish future expansion?

If multiple specs can be active at once in the future, the "elite" field could also just return an array of spec strings that are active.

Any additional thoughts to the above?

rswiderski:
As I think, the main request for your "elite" property should be having properties that we could use to know a specialization/trait/spell is unlock.
I can be interresting to know even for character that aren't 80 if they have unlock this or that spells/traits.

milneti:
So, you can already find out the active traits and specializations through the specializations object in the characters endpoint. The purpose of my change is to give information on the active specialization as it pertains to how a character is identified in the game (I.e. elementalist vs tempest).

rswiderski:
```json
  "specializations": {
    "pve": [
      {
        "id": 15,
        "traits": [
          1776,
          1802,
          1715
        ]
      },
      {
        "id": 9,
        "traits": [
          1810,
          1740,
          1770
        ]
      },
      {
        "id": 52,
        "traits": [
          1716,
          1743,
          1746
        ]
      }
    ],
    "pvp": [
      null,
      null,
      null
    ],
    "wvw": [
      {
        "id": 3,
        "traits": [
          1784,
          1760,
          1749
        ]
      },
      {
        "id": 15,
        "traits": [
          1776,
          1802,
          1754
        ]
      },
      {
        "id": 9,
        "traits": [
          1728,
          1740,
          1770
        ]
      }
    ]
  },
```

You already have it here, on character endpoint just had to combined this specialization endpoint that can say you if a specialization is an elite one

madou:
@rswiderski You're still missing the point. Yes, we can dig through the specializations array to figure out a characters elite class, but Milneti's suggestion is simply a quality of life change to add it to the top level characters resource.

Unless your point is saying you'd like to know which element of the array is elite by a flag? If that's the case it is **always** the third element.

I still suggest to simply change the profession property if a character has selected an elite spec. So for example Necromancer's profession would be 'Reaper' instead of 'Necromancer'. Simple, clean, and it makes sense.

rswiderski:
@madou Forget my first answer, I was only a suggestion. Your suggestion isn't as simple as it seems, how do you manage the fact that a character have three specialization trees? How can you priorize PvE or PvP or WvW specialization to override the profession's name.

You have to keep some form of abstraction on data cause not everybody use it in the same way.

madou:
That's a good point that I've overlooked! Tough sell to change it from what it is atm then.

milneti:
I see having to keep track of different game modes if being able to multi-spec elite specializations could be hard. I guess that would mean one tracker per game mode... I still don't think sifting through that monstrosity of a specializations dump to find out if I should show a character as a reaper or not is a good alternative..

Maybe having it like this would still give an easier way to know and allow the splitting between game modes:
 "elite":[
      {
        "pve": null
      },
      {
        "pvp": "other"
      },
      {
        "wvw": "tempest"
      }
]

However, I don't even know if you would be able to have a different one per game mode? maybe someone from Anet can elaborate toward future plans with this..

I think that the mastery level and map completion percent pieces would be a good addition even still.

lye:
> However, I don't even know if you would be able to have a different one per game mode?

You can currently have an elite specialization equipped in one game mode (e.g., PvP) but not another. Thankfully we fixed the bug where it would unequip your specialization-specific weapon when switching modes; that was super annoying.

milneti:
Ha very true statement... I would think that the change I most previously commented would then account for the different modes as well as the possibility of multiple elite spec options. This would still make it much easier to access the information without having to filter through the whole specializations dump.

I can change the pull request if that sounds like a good future-proofing fix?

milneti:
Modified changes to account for future possibilities.

milneti:
any updates for this? specifically exposing the currently equipped elite spec or exposing map completion %?

lye:
I don't have any updates; exposing the training track progress it probably something I'll tackle within the next month though. One of the prerequisites was exposing the total (and localized) training track info, and that's recently been turned on (via [/v2/professions/](https://api.guildwars2.com/v2/professions/Guardian)).

PR #132, opened 2015-12-07
Add an example of rendering a guild emblem.

### LINKED ISSUE #65: The secrets of rendering guild emblems
So there's the ```v1/guild_details``` API that enables us to render guild emblems on the fly - the question still is: _how?_
The results have been mediocre at best ([chance hit](http://gw2.chillerlan.net/examples/gw2emblems.php?guild_id=75FD83CF-0C45-4834-BC4C-097F93A487AF)) because none of us knows the secrets of postprocessing these images.

So would it be possible for you to give us an example how to properly render guild emblems?
(preferably in a language thats read- and/or usable for most of us (PHP! :P)
The example above has been created by [this cheap examply example code](https://github.com/codemasher/gw2api-tools/blob/master/examples/gw2emblems.php).

See also this thread: https://forum-en.guildwars2.com/forum/community/api/API-Suggestion-Guilds/page/2#post2155863

lye:
Fixes #65

### DIFF
--- a/v2/emblems/example.html
+++ b/v2/emblems/example.html
@@ -1,0 +1,240 @@
+<!DOCTYPE html>
+<style>
+    body {
+        background-color: #eee;
+        text-align: center;
+        padding-top: 20px;
+    }
+    
+    #emblem {
+        width: 128px;
+        height: 128px;
+        border: 1px solid black;
+    }
+    
+    #scratch {
+        width: 128px;
+        height: 128px;
+        border: 1px solid red;
+        visibility: hidden;
+    }
+</style>
+<canvas id="emblem"></canvas>
+<canvas id="scratch"></canvas>
+<script>
+window.onload = function() {
+    var baseUrl  = "https://api.guildwars2.com",
+        guildId = "4BBB52AA-D768-4FC6-8EDE-C299F2822F0F",
+        steps;
+    
+    function getJSON(url, cb) {
+        var xhr = new XMLHttpRequest();
+        
+        xhr.onreadystatechange = function() {
+            if(xhr.readyState !== 4) {
+                return;
+            }
+            
+            return cb(JSON.parse(xhr.responseText));
+        };
+        
+        xhr.open("GET", baseUrl + url, true);
+        xhr.send();
+    }
+    
+    return getJSON("/v1/guild_details.json?guild_id=" + guildId, function(guild) {
+        var data = {
+                guild      : guild,
+                colors     : null,
+                background : null,
+                foreground : null
+            },
+            colorIds = [
+                guild.emblem.background_color_id,
+                guild.emblem.foreground_primary_color_id,
+                guild.emblem.foreground_secondary_color_id
+            ];
+        
+        function checkPromise() {
+            var missing = Object.keys(data).some(function(x) {
+                    return !data[x];
+                });
+            
+            if(!missing) {
+                loadImages(data);
+            }
+        }
+        
+        // XXX: You really want to use promises here, but w/e this is super-vanilla.js.
+        getJSON("/v2/colors?ids=" + colorIds.join(","), function(colors) {
+            data.colors = colors.reduce(
+                function(a, x) {
+                    a[x.id] = {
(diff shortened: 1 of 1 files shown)


darthmaim:
This doesn't look right (ignore the tiled background in your example, was just copying that in to have similar backgrounds):

| Ingame | Exampe |
|-----|------|
| ingame | example |

codemasher:
Another example: https://twitter.com/codemasher/status/673960987310051328

Aralicia:
My guild may be one of the most extreme case. The colors aren't even remotely the same.
orpa_logo
Blue became green, and yellow became bordeaux.

lye:
> My guild may be one of the most extreme case. The colors aren't even remotely the same.

Could you post your `guild_id`? I'm digging deeper into the image compositing code -- it's a bit weird since they're doing transforms in the HSV colorspace. One of the bits that I missed is an implicit brightening; I'm guessing there's a missing hue rotation that's causing the discoloration of your guild icon.

Aralicia:
Guild emblem ID : 810D6274-18D5-4107-80AB-5E974430BB5
There a mistake in colorshift(), within the for(). channel order should be "r g b". Is actually "r b g"
New look, far better : 
orpa_logo

darthmaim:
@Aralicia that's not a valid guild id, it's missing a character at the end: https://api.guildwars2.com/v1/guild_details.json?guild_id=810D6274-18D5-4107-80AB-5E974430BB5

Aralicia:
Woops, missed the last "1" while c/c the ID. fixed in the previous post.

On another note : for the foreground, seems like the red channel of layer 0 serves as a brightness ratio while layers 1&2 serves as alpha masks.

lye:
> for the foreground, seems like the red channel of layer 0 serves as a brightness ratio 

That seems really reasonable. With the B/G fix it looks like the brightness is just off.

codemasher:
And that's where the problems start - we've been there before :D 

https://gist.github.com/codemasher/1cf2a5b0c323b0e67aef

lye:
Yeah, the color matrix stuff is really beyond me. I'm reading what the client does and ... it makes sense syntactically but nothing's really sticking. I'm gonna put up the minor changes I made to the example and throw in a note with that link; pretty sure I'm not going to get much further in any reasonable amount of time.

codemasher:
I think the important part would be [this one](https://gist.github.com/codemasher/1cf2a5b0c323b0e67aef#file-guildemblems-php-L333) where the actual colorize happens which is about the same point where your result is now.

lye:
> But there might also the shaders apply, right?

Nope, the texture is generated entirely on the CPU. There's some lighting effects and such when it's rendered on a model, but when it's displayed on the UI it's just the vanilla texture. The main issue is the colorspace transforms (brightness and constrast) against the HSL values don't match what's being done to generate the RGB tuple from `/v2/colors`. I'm not sure *why* they're using different transforms or exactly how the transforms differ, but that's my best guess.

lye:
Hmm. Does anyone use the RGB pairs from `/v2/colors` for anything? I tweaked the transform on my local stack a bit and got these results:

logo-side-by-side

From left-right: in-game screenshot, tweaked `/v2/color` values, current state. I'm almost thinking that the transforms we've been using in `/v2/color`'s RGB values are just wrong -- the functions used to process the emblems aren't emblem-specific, they're presumably used for all colorizations. Thoughts?

darthmaim:
> Does anyone use the RGB pairs from /v2/colors for anything

The only place I know of are item databases and similar like this: https://en.gw2treasures.com/item/20714, nothing crititcal.

Your tweaked example looks definitely better than before.

lye:
Ugh, I think the changes I made locally are wrong. The cloth material, as served currently, matches the color displayed in the UI (which is probably the right thing). The altered `/v2/colors` (on the right) just look wrong.

nope-colors

I think the correct approach is to use the old transforms, but I've spent way too much time on this today and need to work on other API things. If someone wants to submit a pull request to update the example to output better images, that would be lovely  ;__;

PR #137, opened 2015-12-08
/v2/achievements should probably expose rewards.

lye:
Note that this explicitly excludes titles -- probably want to have an
endpoint that serves localized title strings (since not all of them are
awarded from achievements (or are they?))

### DIFF
--- a/v2/achievements/index.js
+++ b/v2/achievements/index.js
@@ -25,6 +25,16 @@
 			"count"  : 100,
 			"points" : 5
 		}
+	],
+	"rewards"     : [
+		{
+			"type" : "Item",
+			"id"   : 30001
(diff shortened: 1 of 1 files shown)


Aralicia:
There are a few titles without achievements : the "Achiever" series, awarded whenever the account reach an amount of AP; the "of the arena" series, awarded for being in the top 5/10/etc in some PvP Tournaments; and "Maguuma Trailblaizer", award for pre-purchasing HoT.
Theses titles may be associated to hidden achievements, though.

lye:
Will probably add in a `/v2/titles` endpoint (and `/v2/account/titles`) or something at a later point, I think.

PR #140, opened 2015-12-10
Add in /v2/guild/:id/members.

lye:
It lists guild memberships, basically.

### DIFF
--- a/v2/guild/members.js
+++ b/v2/guild/members.js
@@ -11,7 +11,7 @@
 	{
 		name: "Lawton Campbell.1450",
 		rank: "Member",
-		joined: "2015-06-17T02:38:27.000Z"
+		joined: null
 	}
 ]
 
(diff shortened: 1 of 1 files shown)


Teranas:
Is it correct that the only way to determine if a users token is permitted to request the new guild endpoints by trying it out? (Besides checking if the token contains the "guilds" permission flag)

lye:
> Is it correct that the only way to determine if a users token is permitted to request the new guild endpoints is trying it out?

Right now, yeah. You'll get back `{"text" : "access restricted to guild leaders"}` if the API key doesn't belong to a leader, or `{"text" : "membership required"}` if they're not in the guild at all.

Plitzz:
Hello there,

the members is definitely an error in the API:

Joined is clearly in the wrong time zone or the time zone has not been well documented.
I can understand this exactly two guild members:

"joined": "2015-12-24T02: 27: 21.000Z"
On Christmas Eve, no one has joined with us to security.

"joined": "2016-01-04T03: 22: 56.000Z"
Again, the true date / time does not, accession took place approximately at 19:00 CET.

greeting
Plitzz

lye:
@Plitzz -- I created issue #171 to track that. Thanks for the report.

Issue #138, opened 2015-12-10
Missing information to calculate total AP / missing achievements


queicherius:
One of my users with a lot of AP was so friendly to [compile a list](https://www.reddit.com/r/gw2efficiency/comments/3w5s6f/list_of_achievement_ui_bugs/) and was ok with me crossposting them here. 

- Daily / monthly AP (and daily completion status) are missing
- No markers for repeatable achievements
- Missing achievements:
	- Hero achievements (1st, 2nd and 3rd birthday achievement missing)
	- Hall of Monuments ("God walking amongst mere mortals" achievement missing)
	- Tangled Depths (4 achievements missing, the 4 hidden ones)
	- Basic Collections (41 of 68 in the API)
	- Rare Collections (9 of 16 in the API)
	- Black Lion Collections (0 of 27 in the API)
	- Legendary Weapons (23 of 72 in the API)
	- Edge of the Mist (15 of 20 in the API)
	- Boss Week (6 of 7 in the API)
	- Retired Achievements (3 of 6 in the API)
	- 6 out of 37 historical achievement categories missing entirely (Blood and Madness 2013 and 2014, A very marry Wintersday 2013 and 2014 and Shadow of the Mad King 2012 and 2015)
	- Escape from Lion's Arch (15 out of 20 in the API, the 5 repeatables are missing)
	- Queen Jennah's Jubilee (13 out of 14 in the API, Opening Ceremony Attendee missing)
	- Support Evon Gnashblade (11 out of 12 in the API, repeatable minigame achievements missing)
	- Support Ellen Kiel (11 out of 12 in the API, repeatable minigame achievements missing)

A lot of the missing achievements are the one's which are hidden until unlocked.

lye:
> Daily / monthly AP (and daily completion status) are missing

Hmm, I'm not sure where those are stored, actually. Will have to investigate -- they're probably just a counter somewhere. It may make sense to throw an object onto `/v2/account`'s response that has an AP overview (and includes daily/monthly counts).

> No markers for repeatable achievements

Ick, I'll have to add those in. The repeatable achievements are also missing a prerequisite achievement which would be useful with the data.

> Missing achievements

Most of those should be sorted out sometime after the next release. I added in a whitelist system (similar to the item/skin/recipe/etc bits) which watches the logs for achievement progress events and persists the list of achievement ids seen. That way we can track which achievements are actually acquirable and avoid dumping a bunch of debug/testing/in-dev achievements onto the endpoint.

[...]

PR #142, opened 2015-12-10
Add in support for repeatable achievements.

lye:
refs #138 

This is mostly how I'm imagining the output format for these. Not sure when I'll have this ready; going to need some additional backend support to piece together the data -- won't be able to get anything live until the release after the next release.

### DIFF
--- a/v2/account/achievements.js
+++ b/v2/account/achievements.js
@@ -19,6 +19,12 @@
 	"current": 11,
 	"max": 12,
 	"done": false
+}, {
+	"id": 129,
+	"current": 11,
+	"max": 200,
+	"done": false,
+	"repeated": 35
 }]
 
 // "id" references /v2/achievements.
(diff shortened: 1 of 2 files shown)


queicherius:
I am not sure if I am missunderstanding the format or the "how many repeats do give points" information is missing.

Let's say "Hobby Dungeon Explorer" would look like this:

```js
{
	"id" : 123,
	"icon" : "https://render.guildwars2.com/...",
	"name" : "Hobby Dungeon Explorer",
	"description" : "Asdf.",
	"requirement" : "You must asdf.",
	"type"        : "Default",
	"flags"       : [
		"Pvp",
		"Repeatable"
	],
	"tiers"       : [
		{
			"count"  : 5,
			"points" : 5
		} // 5 points per 5 dungeon completions
	],
	"prerequisites" : [
		456 // ID of dungeon master
	]
}
```

And my completion status of that achievement like this:

```js
{
	"id": 123,
	"current": 3,
	"max": 5,
	"done": false,
	"repeated": 346
}
```

Now I know that I am on 3 of 5 dungeons I need to complete to get another 5 points. But I already completed it 346 times and it only gives points for the first 40. That information is missing.

lye:
D'oh, good point. I stuck the extra field in there as `point_cap`, since that's how we're representing it on the content side (rather than a maximum number of completions).

PR #143, opened 2015-12-11
Add bit information to /v2/achievements

Eearslya:
I noticed this was already live, but without any documentation, so I decided to write it up so others could be made aware too. Here's hoping I didn't miss any values for `bits.n.type`.


lye:
Whoops, I totally neglected to do this. Thanks! <3

Issue #121, opened 2015-11-22
Guild treasury?


evaryont:
How feasible is it expose the information from the guild treasuries that a user is part of? For instance, I had thought of making a flax tracker to hopefully better motivate guildies into donating and/or collecting a lot of flax. (Gotta get that linseed oil!) Note, this is different than what's in the guild vault.

I'm aware of #7, but as it was written before guild halls were shipped so it's missing some information that would be useful now. Myself, I'd be looking for a few pieces of information: What upgrades are currently being tracked (say, [Guild Portal](http://wiki.guildwars2.com/wiki/Guild_Portal) or [Guild Enhancement: Experience](http://wiki.guildwars2.com/wiki/Guild_Enhancement:_Experience)), what the costs are for each upgrade, and how many items are already stored in the treasury.

All of this information is already visible in the client. Nor does it seem like it's too privacy sensitive, after authentication to prove you are in the guild. Thoughts?

PR #149, opened 2015-12-22
Add /v2/guild/:id/treasury which dumps treasury contents.

lye:
Basically just dumps out an array of itemstacks that represents the current treasury state. For tracking who donated what (what when), I want to also expose the guild's event log (that's visible in the in-game history tab of the guild panel), but that's a separate pull request.

refs #121

Thoughts on the output format?

### DIFF
--- a/v2/guild/treasury.js
+++ b/v2/guild/treasury.js
@@ -1,0 +1,18 @@
+// Authenticated endpoint that outputs guild treasury.
+// **only available with keys from guild leaders for now**
+// Requires scope: "guilds"
+
+// GET /v2/guilds/:id/treasury?access_token=:token
+
+[
+	{
+		id    : 12138,
+		count : 200
+	},
+	{
+		id    : 19684,
+		count : 3493
+	}
+]
+
+// "id" is an item id and can be resolved against /v2/items.
--- a/v2/guild/treasury.js
+++ b/v2/guild/treasury.js
@@ -6,13 +6,33 @@
 
 [
 	{
-		id    : 12138,
-		count : 200
+		id        : 12138,
+		count     : 200,
+		needed_by : [
+			{
+				upgrade_id : 42,
+				count      : 1000
+			}
+		]
 	},
 	{
-		id    : 19684,
-		count : 3493
+		id        : 19684,
+		count     : 3493,
+		needed_by : [
+			{
+				upgrade_id : 42,
+				count      : 3000
+			},
+			{
+				upgrade_id : 43,
+				count      : 3000
+			}
+		]
 	}
 ]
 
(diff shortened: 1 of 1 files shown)


tivac:
I like the format, but folks actually building against this may have other opinions.

Eearslya:
Is it possible to also include the total amount needed at the time?

queicherius:
> Thoughts on the output format?

:+1: Looks fine, and it's consistent with [/v2/bank](https://github.com/arenanet/api-cdi/blob/master/v2/account/bank.js) (and others)

lye:
> Is it possible to also include the total amount needed at the time?

I was thinking about this and I'm kind of on the hedge. To get the total amount (either on our side or the application's side), you'd fetch the current pending upgrades, then resolve them against `/v2/guild/upgrades` to get the bill of materials, then figure out the maximum of each material and insert it into the appropriate fields. The argument against doing it on our side is that if you're displaying the data in an application, you'd want the localized upgrade names (and possibly icons) anyway so you'd be fetching all of the data to compute the max.

If we did it on the API side the applications would end up fetching all of the data we saved anyway.

Instead of including the max (or, in addition to including the max), we could add a `used_by` field that included both the upgrade id and the quantity needed by the upgrade, e.g.

```json
{
    id : 42, // item id
    count : 100 // how many are in the treasury
    max : 3000 // used_by.reduce(function(a, x) { return a + x.count }, 0)
    used_by : [
        {
            id : 99, // upgrade id
            count : 200 // number wanted by this upgrade
        },
        ...
    ]
}
```

The main incentive behind this format is that the application can fetch only the upgrade data that's needed to display which upgrades need stuff from the treasury (as well as just providing the max value if you don't care about displaying the upgrades -- or want to defer displaying them until the user requests an expanded view).

Not really fond of the name `used_by` though, eh.

Eearslya:
Seems like a fair idea to me; maybe `needed_in` or `needed_by` instead as a
name?

lye:
`needed_by` sounds good. Also dropped the `max` field since I don't think including it really adds anything (and I don't want to add too many extra bits and bobs to an otherwise normal itemstack).

Eearslya:
Shockingly, I don't think this has been discussed anywhere, but..Do you think this might be a good place to put aetherium too? As far as I can see it hasn't been discussed anywhere else just yet, and it seems a little silly to have a full-fledged `/v2/guild/:id/aetherium` endpoint.

lye:
I was thinking about sticking aetherium and stuff into `/v2/guild`, actually. Haven't really decided what I'm going to do with that yet.

Eearslya:
Fair enough; I hadn't thought about V2's generic `guild` endpoint. I do agree though, that's probably a better place.

PR #150, opened 2015-12-22
Add the first bits of /v2/guild/:id/log.

lye:
This exposes the data used by the in-game guild history panel. Starting with a small subset (due to some additional work needed to retrieve other events) which only includes `treasury` and `motd` events. We'll eventually expand this to include membership changes, missions, etc.

refs #121

### DIFF
--- a/v2/guild/log.js
+++ b/v2/guild/log.js
@@ -4,22 +4,70 @@
 
 // GET /v2/guilds/:id/log?access_token=:token
 
+// Example roster events.
 [{
-	"id": 1190,
-	"time": "2015-12-10T23:58:49.106Z",
-	"type": "treasury",
-	"user": "Lawton Campbell.9413",
-	"item_id": 24299,
-	"count": 150
+	"id": 1377,
+	"time": "2016-02-16T01:08:10.539Z",
+	"type": "joined",
+	"user": "Lawton Campbell III.7615"
 }, {
+	"id": 1375,
+	"time": "2016-02-16T01:08:07.538Z",
+	"type": "invited",
+	"user": "Lawton Campbell III.7615",
+	"invited_by": "Lawton Campbell.9413"
+}, {
+	"id": 1217,
+	"time": "2015-12-16T01:18:46.536Z",
+	"type": "kick",
+	"user": "Lawton Campbell II.8307",
+	"kicked_by": "Lawton Campbell.9413"
+}, {
+	"id": 1214,
+	"time": "2015-12-16T01:18:25.534Z",
+	"type": "rank_change",
+	"user": "Lawton Campbell II.8307",
+	"changed_by": "Lawton Campbell.9413",
+	"old_rank": "Member",
+	"new_rank": "Foo"
+},
+
+// Example MOTD events.
+{
 	"id": 1111,
-	"time": "2015-12-10T23:52:01.099Z",
+	"time": "2015-12-10T23:52:01.493Z",
 	"type": "motd",
 	"user": "Lawton Campbell.9413",
 	"motd": "FSFOSDFJS"
+},
+
+// Example treasury events.
+{
+	"id": 1283,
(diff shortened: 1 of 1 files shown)


Eearslya:
Just to clarify something for me, will this endpoint just spit out everything it can at once, all on one page?

lye:
Yes. Due to our log retention policies, the dataset size is fixed at ~100 entries/type, so the gzip'd size is going to be fairly manageable to send in a single response.

hackedd:
You note above that only `treasury` and `motd` events are supposed to be available, but I'm seeing some others as well:

* inf_combined
* influence
* invited
* joined
* kick
* rank_change
* upgrade

Is there a list of all types, what they mean, and what fields to expect?

Is there a reason that `inf_combined` has a `created` timestamp, while all other types I've seen so far have a `time`?

lye:
Forgot to merge, derp.

> Is there a list of all types, what they mean, and what fields to expect?

Uhh, I thought I had put all the documentation into these, but after double-checking the PR ... yeah I should do that. I'll update with another commit before merging.

> Is there a reason that inf_combined has a created timestamp, while all other types I've seen so far have a time?

That's a bug -- IIRC `inf_combined` doesn't have a parser to extract actual values from the `rawData` field (can you see the `rawData` field?) and for some reason is still being omitted. I'll see if I can get that fixed with a deploy later today (after I look into a treasury issue).

hackedd:
The `inf_combined` entries I can see all look like this:

```
{
    "created": "2015-10-22T21:08:27.301Z",
    "id": 2174,
    "type": "inf_combined"
}
{
    "created": "2015-10-21T21:06:40.301Z",
    "id": 2163,
    "type": "inf_combined"
}
```

lye:
I'll have `inf_combined` events removed later today; I'm debating removing the `influence` events too since they don't make much sense anymore, but haven't decided yet.

loriamor:
There must still be items which give influence as we have had recent entries regarding influence in the guild log.

lye:
I definitely think there are still some lurking about; the usages of those will appear in the `influence` entries (which I'm leaving in for now). The `inf_combined` ones are the summary of total influence earned/day and thus aren't very useful, I think.

Sigma-:
Is the file guild.py with the treasury working ? I'm about to begin a project and i need to get the items'id from a guild treasury ! Thanks for replying

lye:
What is `guild.py`?

Riekle:
Would it be correct that the treasury is not yet added to this log? It's in the documentation here and on the wiki but I can't seem to get it when pulling data from the API.

hackedd:
I do see `treasury` events in the logs, so there might be a problem with your API key?

```
{
    "id": 199948,
    "time": "2016-04-07T20:36:35.419Z",
    "type": "treasury",
    "user": "...",
    "item_id": 24277,
    "count": 50
},
````

Riekle:
Odd, I'll take a look into the API permissions for that account. Don't see any other reason why it shouldn't show those particular entries.

lye:
It should just require the "guilds" permission -- if you can see the log, you should also be able to see treasury events. It's entirely possible that something's horribly broken, but in most of those cases the endpoint *should* return `{"error" : "endpoint broken"}`.

Riekle:
After some trying it seems that just older reports aren't visible. I've redeposited into the treasury now and I can see it. Can it be that unlike the other logs the treasury logs only remain as long as they would ingame? (Meaning the tons of super clouds bought have made everything else disappear)

lye:
>  Can it be that unlike the other logs the treasury logs only remain as long as they would ingame?

As far as I'm aware, *all* the logs are only around as long as they are in-game. The API is pulling from the same data source as the client. The super cloud situation is a bit unfortunate, but unavoidable. Hopefully the next time something like that's added it'll work with the multiple-purchase feature :|

PR #154, opened 2015-12-24
Add A Way to Distinguish Instances From Normal Maps

FrederikNS:
It is currently rather troublesome to figure out which maps are instances and which are normal maps.

To do anything useful I have resorted to having a blacklist of all the instance maps, but this is tedious, and when more instances are introduced, I have to add the new instances.

I propose that an "instance" boolean is added to the maps, such that they can be clearly distinguished.

### DIFF
--- a/v2/floors/floor.js
+++ b/v2/floors/floor.js
@@ -11,6 +11,7 @@
                     "min_level" : 40,
                     "max_level" : 50,
                     "default_floor" : 1,
+                    "instance": false,
                     "map_rect" : [
                         [-27648, -36864 ],
                         [ 27648, 39936 ]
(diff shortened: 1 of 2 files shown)


R0dolf:
Hi, 
I'd like to know if there's any progress on this subject ?
I also find that it's quite hard to filter "normal" maps from others via the continent API since all instance maps are also there. Actually, the only way to do it is by creating a whitelist or a blacklist so, it's hard to keep it updated.
I think that adding a boolean could be a good idea ?

lye:
Superceded by #333 which is live. Should be able to determine whether a map is an instance via `map.type === "Instance"`.

Issue #156, opened 2015-12-25
Skin rarity


BryghtShadow:
These skins have orange names in-game, but there is no rarity field in the API.
https://api.guildwars2.com/v2/skins?ids=6059,6360,6183,6418,6043,6050,6375,6479,6350

lye:
AFAIK no skins have an associated rarity. Rarity is an item attribute.

lye:
Apparently pre-coffee I am literally blind. There totally is a rarity field, how did I miss that.

PR #163, opened 2016-01-04
Add skin rarities and rarity override flag.

lye:
refs #156

### DIFF
--- a/v2/skins.js
+++ b/v2/skins.js
@@ -16,6 +16,7 @@
 	"flags" : [
 		"ShowInWardrobe"
 	],
+	"rarity" : "Basic",
 	"restrictions" : [],
 	"details"      : {
 		"type"         : "Leggings",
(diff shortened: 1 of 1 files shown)


v2/skins.js
```diff
@@ -80,6 +84,7 @@
 //   * "ShowInWardrobe"
 //   * "HideIfLocked" -- doesn't show up in the wardrobe if not unlocked
 //   * "NoCost" -- doesn't cost transmutation charges to apply
+//   * "OverrideRarity" -- skin overrides item rarity when applied
```
Fishrock123: Could this be clarified a bit more? I don't understand what it is supposed to do.

lye: e.g., if you apply a legendary skin to an item, it changes the item's border to legendary.

Pretty sure that's what the flag does.

PR #178, opened 2016-01-15
add utility skills for character builds info /v2/characters

degger80:


degger80:
for revenants should be included choosen legendary stances
```
stances: [123,345]
```
and for rangers - choosen pets
```
pets: [123,345]
```

lye:
Hmm, I was actually just thinking about this on the way in. It's a bit weird since pets/stances (IIRC) aren't just skills -- there's another data structure for each of 'em. Will probably want an endpoint that dumps out all of the pets and another one that dumps out all of the stances so they can be referenced by `pet_id` and `stance_id`.

I'm not sure what else would go in i.e. `/v2/stances` but the pet endpoint I think could have stats and which skills the pet could use. Need to double-check what data there is; I typically don't play professions that start with `r`.

degger80:
Well, revenant stances are skills:  
https://api.guildwars2.com/v2/skills/27659
27659 - Legendary Assassin Stance
So they already can be included to build info without any additional endpoints for stances.

What about rangers pets... I agree - it can be implemented only after pets endpoint completed.

lye:
I'll adjust it when I merge this, but I renamed `slot_skills` to just `skills`. Also `pets` are there.

degger80:
may be stances also should be separated from skills. like pets now.

Issue #174, opened 2016-01-14
Scribe recipes with missing items from ingredients


BryghtShadow:
There are scribe recipes that are built using items in guild storage (namely banners and decorations). However, these items are missing from the `ingredients` array.
* Out of 408 Scribe recipes, there are 176 recipes that are built from items in guild storage: https://api.guildwars2.com/v2/recipes?ids=9917,9919,9926,9928,9938,9941,9959,9975,9979,9983,9985,10004,10006,10008,10014,10030,10037,10038,10049,10060,10071,10072,10076,10081,10084,10087,10097,10101,10103,10106,10113,10129,10142,10152,10168,10180,10183,10199,10217,10237,10257,10258,10260,10263,10265,10293,10294,10356,10357,10360,10371,10383,10391,10395,10402,10412,10423,10431,10434,10461,10462,10470,10475,10479,10482,10499,10510,10521,10530,10557,10565,10566,10567,10579,10599,10601,10622,10626,10635,10643,10646,10669,10686,10691,10693,10714,10765,10785,10790,10801,10810,10812,10821,10856,10867,10870,10887,10893,10921,10928,10965,10968,10977,10994,11009,11022,11024,11030,11031,11086,11099,11103,11113,11119,11135,11145,11148,11149,11151,11163,11174,11185,11186,11220,11246,11257,11261,11287,11301,11306,11318,11319,11320,11323,11326,11337,11339,11345,11374,11383,11430,11452,11457,11460,11469,11475,11489,11495,11503,11511,11513,11528,11530,11537,11538,11563,11582,11591,11594,11609,11635,11636,11637,11645,11672,11674,11691,11702,11709,11727,11734,11746,11750,11752,11753,11754
* Out of the above 176 recipes, 172 have type `"GuildDecoration"`; the other 4 (`[10168,11009,11185,11563]`) have type `"GuildConsumable"`.

## Example
https://api.guildwars2.com/v2/recipes/10183
```json
{
  "type": "GuildDecoration",
  "output_item_id": 71716,
  "output_item_count": 1,
  "min_rating": 125,
  "time_to_craft_ms": 1000,
  "disciplines": [ "Scribe" ],
  "flags": [ "AutoLearned" ],
  "ingredients": [
    { "item_id": 70454, "count": 1 },
    { "item_id": 75087, "count": 1 },
    { "item_id": 19713, "count": 5 }
  ],
  "id": 10183,
  "chat_link": "[&CccnAAA=]"
}
```
[...]

Issue #152, opened 2015-12-23
Scribe recipes that don't produce items don't go through the whitelist.


lye:
I'm guessing the output type isn't "Item" or something.

([forum thread](https://forum-en.guildwars2.com/forum/community/api/Items-mentioned-in-recipes-not-found))

lye:
So this is kinda weird -- scribe recipes that produce a guild upgrade rather than an item *still* have an "output item". The output item never gets given to the player, however, so it never gets put in the whitelist queue. #189 should fix the issue w.r.t not being able to tell which recipe produces which output (since you'll be able to look up the name via `/v2/guild/upgrades`). In these cases I'd *like* to remove the `output_item_id` since it's meaningless, but since it's intuitively a "required" field, removing it will probably break stuff.

In any case, the spirit of the issue (being unable to tell which recipes produce which items -- turns out this has nothing to do with whitelist issues) should be fixed by #189.

PR #189, opened 2016-01-19
Add in guild_ingredients and output_upgrade_id.

lye:
These refer to the scribe recipe components that are pulled from/put into
the guild inventory (because they're not items -- they're actually guild
upgrades, e.g., decorations or schematics).

refs #174

### DIFF
--- a/v2/recipes.js
+++ b/v2/recipes.js
@@ -22,8 +22,8 @@
 	"chat_link": "[&CQEAAAA=]"
 }
 
-// GET /v2/recipes?ids=1
-// GET /v2/recipes?page=0&page_size=1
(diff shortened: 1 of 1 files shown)


lye:
Going to copy some notes from #152 here for clarity: the `output_item_id` in Scribe recipes that produce a guild upgrade is totally meaningless. I would like to remove it, but suspect removing it will break a bunch of stuff.

AFAIK, all current scribe recipes have an `output_item_id` but this is not enforced in content and future recipes could potentially have this field omitted (if there isn't one already). Once that forces my hand I'll feel more comfortable removing the field.

Issue #151, opened 2015-12-22
Need to view current/pending guild upgrades.


lye:
Needed for computing the treasury max values (and for, you know, viewing which upgrades a guild has).

refs #134

queicherius:
What are you including into "pending" guild upgrades? The scribe crafting things?

lye:
I was thinking that the scribe-created schematics and such should go into a separate endpoint -- I was going to focus only `Unlock`-type (and I guess `AccumulatingCurrency`; I'm not sure why those are distinct) upgrades for this endpoint. I'm mostly looking to expose the tech tree (which contains upgrades that can be acquired exactly once) rather than the consumable bits (which should have a `count` field and a separate endpoint, IMO).

queicherius:
> I'm mostly looking to expose the tech tree (which contains upgrades that can be acquired exactly once) rather than the consumable bits (which should have a count field and a separate endpoint, IMO).

That's what i'd love to see! :+1:

My question originally came up because I think these things only have "completed" / "not completed" but no "pending" status. Or am I wrong with that?

lye:
Hmm, I might be mis-understanding, but I thought for the items to appear in the treasury the upgrade had to be manually selected or something; then once the requisite items were deposited you could complete the upgrade.

Needless to say I'm probably misunderstanding the system.

queicherius:
> Hmm, I might be mis-understanding, but I thought for the items to appear in the treasury the upgrade had to be manually selected or something

Only the prerequisites (upgrade prerequisites, e.g. "Tavern Restoration 1" for "Mine Excavation 1", and guild level) have to be met for this. When you could theoretically build them you are able to deposit items. Then when you have the appropriate amount you can click "upgrade" and it completes instantly.

lye:
Deployed!

PR #193, opened 2016-01-21
Add /v2/guild/:id/upgrades.

lye:
Authenticated endpoint for retrieving the completed upgrades for a guild.

The consumable/decoration stacks, though technically upgrades, will eventually go into `/v2/guild/:id/inventory` or something. They have their own endpoint because they're going to use an itemstack output format rather than a list of ids.


lye:
refs #151

PR #198, opened 2016-01-27
Break /v2/commerce/listings to not return the entire set.

lye:
Currently /v2/commerce/listings returns potentially huge responses which
are only useful for a small subset of applications -- but a small subset
of the data returned is valuable more widely. This incongruity is causing
some undue stress; the goal of this breaking change is to limit the
response size for the common case while still providing access to the full
dataset for applications which need it.

By applying pagination.

In the current revision, aggregates are provided in all responses -- not
sure how much I like this (or the aggregate field names). This should make it
so that applications wishing to display the total number of listings/quantity
don't have to fetch the whole dataset.

Anyway, this is a pretty big breaking change (in that applications wanting the
entire result set will need to be updated). I don't have a timeline yet but spread
the word.

### DIFF
--- a/v2/commerce/listings.js
+++ b/v2/commerce/listings.js
@@ -1,28 +1,56 @@
-// Normal bulk-expanded endpoint that exposes detailed listing information
-// about items on the Trading Post.
+// Endpoint that exposes detailed listing information
+// about item listings on the Trading Post.
 
 // GET /v2/commerce/listings
 [ 1, 2, 3, 4, ... ]
 
 // GET /v2/commerce/listings/5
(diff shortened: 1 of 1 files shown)


lye:
Hmm, pulled the entire dataset on live and it's only ~30MB -- going to look at caching the entire dataset locally in each API server (rather than using the distributed cache). That might reduce the load enough on our network for the problem to be moot.

gratuitous image

lye:
download1

Was able to fix it without breaking anything. Hurrah.

JoluMarti:
Yay! I was scared

Issue #186, opened 2016-01-18
Add /v2/pvp/leagues (and add standing info to /v2/pvp/stats).


lye:
I need to look into this one a bit more to see what we can expose -- leagues are mostly data-driven (e.g., each season has a set of divisions with their own separate rulesets and such, all encoded in content) but I'm not sure if/how we're tracking historical account-league standings. If we are storing historical standings (which would make sense), I'm not sure it makes sense to put it into `/v2/pvp/stats` -- it might make sense to have a separate endpoint for fetching the account's standing for a given league.

Thoughts?

lye:
All the bits should be deployed.

PR #199, opened 2016-01-27
/v2/pvp/seasons and additions to /v2/pvp/games.

lye:
Basically just some endpoints that dump out basic information about various PvP seasons, and in which season a game belongs. I'll make another PR later which contains authenticated season standing data (e.g., maybe by adding it to `/v2/pvp/stats`) but for now this just associates games with seasons, where appropriate (and also exposes which queue the game was played in).


lye:
refs #186

PR #229, opened 2016-02-22
Add flags to /v2/account to call out special accounts.

lye:
Need to root around and see if there are other flags that should be exposed, but `SpecialEvent` is the one I've seen so far.


lye:
We ended up just deleting all the accounts that had the `SpecialEvent` flag (and then deleting that flag), so I'm gonna close this out without merging.

PR #257, opened 2016-03-10
Add guild details

Ruhrpottpatriot:
Currently guild details is the only v1 endpoint left. This PR tries to move the endpoint to v2 of the api.

Problems this PR tries to solve:
1. Guild details were only available when the user knew the id beforehand
2. Exposing too much sensitive information is not preferable
3. Guilds should be able to advertise themselves

Solutions taken by the PR:
1. By calling v2/guilds the user gets a list of all current guilds in the game. Without any suitable endpoint or authentication scope this information will be useless and not harmful
2. By calling v2/guilds/{id} or v2/guilds?ids={ids} the user will get a list of _public_ guild information. This information centres around information a guild would willingly give to the piblic for the sake of advertising. This means we expose Name, Tag, Current and Max Member Count as well as the emblem.

Possible Solutions not taken by this PR:
1. Add a recruiting text for the outside world. This was intended, but not possible since there is no representation of this in-game.
2. Add a field of all members who can invite new recruits. Possible security implications by exposing member names via a public API spoke against this (maybe add this with authentication).

### DIFF
--- a/v2/guild/details.js
+++ b/v2/guild/details.js
@@ -22,6 +22,8 @@
      "background": 27,
      "foreground": 114,
      "flags": [],
+     "member_count": 20,
+     "member_capacity": 50,
      "background_color": 11,
      "foreground_primary_color": 584,
      "foreground_secondary_color": 64
--- a/v2/guild/details.js
+++ b/v2/guild/details.js
@@ -1,0 +1,32 @@
+// Unauthenticated endpoint which outputs public details of a guild.
+
+// GET /v2/guilds
+// This will output all currently registered guilds ids to the user.
+// Example:
+[
+    "F05AA71F-37CD-4857-B632-517909765E4F",
+    "20034A78-E176-44B5-A5FE-07F80C1E7A05",
+    "3E579C5D-055F-48A6-9E09-F921588F83B1",
+    "61443D40-BACD-4D09-83A6-3C761A7BAAE3",
+    ...
+]
+
+// GET v2/guilds/:id or v2/guilds?ids=:id
+// This will output the public details of a guild a given od
+// Example:
+{
+   "id": "75FD83CF-0C45-4834-BC4C-097F93A487AF",
+   "name": "Veterans Of Lions Arch",
+   "tag": "LA",
+   "emblem": {
+     "background": 27,
+     "foreground": 114,
+     "flags": [],
+     "background_color": 11,
+     "foreground_primary_color": 584,
+     "foreground_secondary_color": 64
+   }
+ }
+
+ // GET v2/guilds?ids=:ids
+ // Same as above only that an array of guild details is returned
--- a/v2/guild/details.js
+++ b/v2/guild/details.js
@@ -22,6 +22,7 @@
      "background": 27,
      "foreground": 114,
      "flags": [],
+     "recruting_text": "Lorem Ipsum"
      "member_count": 20,
      "member_capacity": 50,
      "background_color": 11,
--- a/v2/guild/details.js
+++ b/v2/guild/details.js
@@ -22,7 +22,6 @@
      "background": 27,
      "foreground": 114,
      "flags": [],
-     "recruting_text": "Lorem Ipsum"
      "member_count": 20,
      "member_capacity": 50,
      "background_color": 11,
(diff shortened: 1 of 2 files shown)


lye:
One of the big features we'll need is the ability to resolve a guild by name. I saw in gitter a suggestion to add `/v2/guild/search?name=foo` and that seemed like a reasonable idea to me, especially if `/v2/guild` emits all guilds an account is a member of.

The other bit is that there's per-guild data that I'd like to expose in `/v2/guild/:id` that requires authentication -- things like the current amount of aetherium, guild level, current MOTD, and other related things. We can make it so that `/v2/guild/:id` exposes public data if no key is provided and adds in the extra data if a valid key is included with the request (if the provided key is invalid or not from a guild member I think it should emit an error).

I like the recruiters idea, but definitely want to hold off on that for the initial release.

Thoughts?

Ruhrpottpatriot:
Initial thought was to give out all in-game guilds by `v2/guilds`, not just a users since you can already get them by calling `v2/account`. If `v2/guilds` would only expose a users guild, we could not expand on the recruiting idea later on and we would need an authentication, which would clash with ´v2/guilds/{id}`and `/v2/guild/search?name=foo`.

> We can make it so that /v2/guild/:id exposes public data if no key is provided and adds in the extra data if a valid key is included with the request
Seems like a good solution to me.

> I like the recruiters idea, but definitely want to hold off on that for the initial release.
Only a brainstorming idea. For me recruiting would have to include more than a recruiting message, so moving it to later is fine.

lye:
> Initial thought was to give out all in-game guilds by `v2/guilds`

Unfortunately, this is not technically feasible -- there by design isn't a way to enumerate all guild objects. 

I don't mind having some duplication between `/v2/account` and `/v2/guilds`; neither do I mind leaving `/v2/guilds` empty until we find a use for it. The latter feels more future-proof though.

Ruhrpottpatriot:
Then we should leave v2/guilds empty for future use. When we are looking at recruiting support from the api down the road, there should be the possibility to enumerate all in-game guilds.

> Unfortunately, this is not technically feasible --  there by design isn't a way to enumerate all guild objects.

Quick brainstorm on that one: It should be possible to enumerate all players on your side of the api, right? If so, you should be able to query all player info and generate a set of guilds that way.

Aralicia:
> It should be possible to enumerate all players on your side of the api, right? If so, you should be able to query all player info and generate a set of guilds that way. If this is not possible, we should leave v2/guilds empty for future use.
2 points :
- even if possible, enumerating all player and querying all guilds from them to get a definitive list of guilds would probably be really costly, given the total amount of accounts.
- every guilds may not actually want to be so freely accessible. On the top of my head, bank storage guilds from pre-HoT accounts would simply clutter such a listing, and private, familial guilds and RP guilds that try to go for the assassins/spies/thieves guild feeling probably won't want to be listed on a external recruitment site.

Proposition : Allow the guildmaster to register (ingame ?) the guild to a public list. Make this list available through the api. Such a proposition need changes that goes further than just api-side, but would probably be more practical.

Ruhrpottpatriot:
The amount of guilds should not change that quickly, thus running the algorithm once or twice per day and caching the result should be enough to get a sufficiently accurate result.

And I'm not proposing to make every detail public. This algorithm would just output the id which you than can further query for already public information. Any private information is tied to an api-key, as @lye already suggested.
As already said: Recruiting is down the road. For now this PR aims to move `v1/guild_details.json` to `v2/guilds` we are not doing anything, you couldn't already do with a bit of ingame tracking (namely: get details from any guild, as long as you know the name and/or id)

lye:
What do you think about changing the format of `emblem`? I'm thinking

```js
{
	background: {
		id: 23,
		colors: [
			673
		]
	},
	foreground: {
		id: 166,
		colors: [
			443,
			4
		]
	},
	flags: []
}
```

lye:
So, when an API key is included in the request I'm returning the following additional fields:

```js
{
    level: 8,
    influence: 990000,
    aetherium: 31,
    resonance: 10800,
    favor: 9575,
    motd: "FSFOSDFJS"
}
```

queicherius:
@lye This PR seems to be done already. :)

tivac:
Agreed!

PR #258, opened 2016-03-10
Add Titles to /v2/achievements.

Prademix:
Would be nice to see Titles rewarded for achievements exposed in the `rewards` section.


lye:
Pulled this into a bigger PR -- made a minor change to have the title reference `/v2/titles` since `/v2/account/titles` needed to reference 'em too (and `/v2/characters` will need a reference too -- I should add that in). Refer to #281

Issue #263, opened 2016-03-15
/v2/professions -- add weapon skills.


lye:
Whoops.

:<

PR #265, opened 2016-03-17
Add .weapons field to /v2/professions.

lye:
Have an implementation as-described, but it'll take a couple weeks for the backend changes to go through the pipeline. Let me know if the output format could be improved (or if there's class mechanics I missed -- will likely add DS/CA skills somewhere else since they're transforms, not weapons) (also for other profession mechanics (shatters, virtues)).

refs #263

### DIFF
--- a/v2/professions.js
+++ b/v2/professions.js
@@ -88,7 +88,7 @@
 				// Each thief dual-wield indicates which
 				// offhand it requires.
 				{
-					"id"      : 14003",
+					"id"      : 14003,
 					"slot"    : "Weapon_3",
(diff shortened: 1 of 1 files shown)


v2/professions.js
```diff
@@ -75,7 +75,40 @@
...
+				// Most skills look like this.
+				{
+					"id"   : 13004,
+					"slot" : "Weapon_1"
+				},
+
+				// Each thief dual-wield indicates which
+				// offhand it requires.
+				{
+					"id"      : 14003",
```
tivac: This shouldn't have a trailing quote

Issue #205, opened 2016-01-28
Add pvp amulets, sigils and runes to /v2/characters.


lye:
IIRC the `builds.pvp` section of `/v2/characters` is pulling data from the pre-unification patch data; it should be updated to expose the amulet and correct runes/sigils.

PR #269, opened 2016-03-24
Add pvpbuild to /v2/characters, add /v2/pvp/amulets.

lye:
Not entirely happy with jamming it onto `.pvpbuild`, but alas.

refs #205

### DIFF
--- a/v2/characters/characters.js
+++ b/v2/characters/characters.js
@@ -278,6 +278,16 @@
 			}
 		]
 	},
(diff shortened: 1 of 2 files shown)


Aralicia:
Maybe  `.equipment_pvp` would be better ?
Ideally, I think that `.equipment.pvp` and `.equipment.pve` would have been the best, but it would be a breaking change, which isn't great.

Issue #145, opened 2015-12-21
Add guild team PvP info.


lye:
Should be able to:

 * Get a list of teams for a guild.
 * Get the aggregate statistics for each guild team (e.g., wins/losses/etc).
 * Get guild team data into `/v2/pvp/matches`.
 * See the guild team leaderboard if there's an active league.

([url=https://forum-en.guildwars2.com/forum/community/api/Request-for-pvp-games-and-pvp-stats/5865373]forum[/url])

Aonwy:
Is that last piece still something desired?

PR #277, opened 2016-04-05
Add season info to /v2/guild/:id/teams.

lye:
The overall season standings will come later; still fiddling with how to best expose them and adventures (which use the same subsystem on the backend).

refs #145


lye:
This should be live.

Issue #268, opened 2016-03-23
Expose titles


queicherius:
It would be cool if we got the ability to work with the titles in the game.

- `/v2/titles` which works like `/v2/achievements` and exposes all available titles
- `/v2/account/titles` which works like `/v2/account/achievements` and lists all titles an account achieved

PR #281, opened 2016-04-12
Add /v2/titles, /v2/account/titles and title rewards to /v2/achievements.

lye:
Supercedes #258 -- the title names are localized, so they need to reference a localized endpoint (this might be a moot point since `/v2/achievements` is already localized -- it's totally possible to put the title name directly in the achievement). It's kind of bikeshedding on my part but I think it's better with the id in there. Let me know if you have an opinion either way -- the backend support for this won't be in until the release after next.

Refs #268.

### DIFF
--- a/v2/account/titles.js
+++ b/v2/account/titles.js
@@ -1,0 +1,10 @@
+// GET /v2/account/titles
+// Authorization: Bearer API-key
+// Requires "account" and "unlocks" scopes.
+[
+	1,
+	2,
+	3
+]
+
+// Returned ids are titles and can be resolved via /v2/titles.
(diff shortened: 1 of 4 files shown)


Prademix:
Using IDs in `/v2/achievements` seems fine, since it makes titles more consistent to item rewards.

Since most but not all titles are unlocked through an achievement `/v2/titels` should specify how a title is unlocked. This may be `achievement_id` -- references `/v2/achievements` -- and `needed_ap` -- gives the AP needed to unlock the titel (this should include all titles not unlocked by an achievement).

lye:
Ah yeah, a back-reference from title->achievement is a good idea. I'll add that in later today.

hackedd:
Minor nitpick, but `/v2/titles` seems to be missing the locale aware flag...

Issue #110, opened 2015-10-31
broken trait ids in /v2/traits


mstampfli:
Hi all,

Found at least one broken traits id in the traits api.

ID 14 / Revenant Trait

Character has currently this id in the spezalizations array, but is not in the traits api.

cheers

lye:
I think these are the hidden minor traits -- the ones that affect whether or not you can equip the elite-only weapons. We're currently excluding them from `/v2/traits` and `/v2/specializations` because they aren't displayed in-game anymore. Will look into removing them from `/v2/characters` as well.

Thanks for the report  <3

Eearslya:
The weapon proficiency minor traits are actually still shown in game, in the corner of the elite spec's icon.

lye:
Ah nuts, they added them back at some point I guess.

I'll figure something out -- I'm thinking about adding them back to `/v2/traits`, then maybe adding an additional field to `/v2/specializations` to indicate they're implicit traits. I don't want to add them back to the `minor_traits` array since ... having four minor traits feels inconsistent?

Thoughts?

Eearslya:
I'm fine with either, personally. Either way they're easily discernable, since the weapon proficiency trait is always tier 0. Not sure what I'd call the new field though. Perhaps `weapon_trait`?

lye:
Okay, this should be sorted. Let me know if further changes are needed!

PR #282, opened 2016-04-13
Add specialization.weapon_trait.

lye:
refs #110 tangentially.


sm-k:
let me get hired

lye:
[http://arena.net/#careers](http://arena.net/#careers)

PR #287, opened 2016-04-19
First version of story progression

andreasg123:
This is a first attempt to describe story progression.  It is modeled somewhat after `/v2/achievements`.

Choices made in the story are represented as flags (e.g., "Commoner" and "Vigil").  Story lines have chapters that correspond to the entries in the story journal.

For a character, one would see the currently selected story mission, the maximum progress in all story lines, and the flags for choices.  One should be able to reconstruct the path through the story by looking at the flags.

### DIFF
--- a/v2/characters/story.js
+++ b/v2/characters/story.js
@@ -1,0 +1,20 @@
+// GET /v2/characters/Hello
+// Authorization: Bearer token123
+// Scopes: account, progression
+
+// story.selected is the currently selected story mission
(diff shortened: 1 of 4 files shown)


lye:
This is actually pretty close to what I've got right now; there's a handful of other bits (specifically, the story choices that you make during character creation) that need to be exposed, but I'll pull this commit into a separate PR once I've got all the fiddly bits functioning.

PR #319, opened 2016-05-20
Add flags to /v2/skills. Only flag is GroundTargeted.

lye:
Will probably be more flags that I've forgotten though. [Forum ref](https://forum-en.guildwars2.com/forum/community/api/Ground-Targeting-Skills).


Eearslya:
`"Underwater"` would be a good flag. Or, more accurately, `"NoUnderwater"` to match how it's represented on the in-game UI.

PR #360, opened 2016-09-14
/v2/backstory/* -- make answer ids strings.

### LINKED ISSUE #352: /v2/backstory/answers: missing mesmer choices
I cannot find the 3 mesmer mask choices in https://api.guildwars2.com/v2/backstory/answers?ids=all
Sample data:
* Kaori Emora: `{"backstory":[53,105,108,157,189]}`
* Illusionary Riqua: `{"backstory":[53,69,72,74,188]}`

lye:
Should fix #352

Apparently answer ids aren't guarenteed to be unique -- the mesmer and
revenant backstory question has the same answer ids (that's the only case
of overlap). To support this, change the answer ids to be the
concatenation of the parent question with the original answer id (so that
mesmer/revenant backstory answers have unique ids).

This is an unfortunate change and is very backwards incompatible, but I
couldn't come up with a better solution :(


v2/backstory/answers.js
```diff
@@ -4,13 +4,13 @@
 
 // GET /v2/backstory/answers
 
-[ 1, 2, 3, /* ... */ ]
+[ "1-1", "1-2", "1-3", /* ... */ ]
 
 // GET /v2/backstory/answers?id=181
 // GET /v2/backstory/answers/181
```
queicherius: These two IDs would need updating too

PR #381, opened 2016-10-04
/v2/wvw/matches: add skirmish data and victory points.

lye:


lye:
The backend work for this totally fell apart on live because I'm terrible.

PR #385, opened 2016-10-23
Added salvage table to items details

fresnizky:
Add salvage table for salvagable items, listing all possible salvages with min values (using crude salvage kit), and max values (using master salvage kit)

### DIFF
--- a/v2/items.js
+++ b/v2/items.js
@@ -24,8 +24,29 @@
     "restrictions": [ ],
     "id": 12138,
     "chat_link": "[&AgFqLwAA]",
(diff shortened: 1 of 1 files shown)


ChieftainAlex:
this sounds like a ridiculous amount of content to fit into an api query.

fresnizky:
That's a good point. It could be made available only when you query for a single item

tivac:
Our system doesn't tend to differentiate between a single item and 200 items, it all looks the same in code. Changing the response based on # of items requested seems like a surprising choice and surprises are real bad in any API you want to consume.

I think it would need to be broken out into a `/v2/salvage` style endpoint where you could cross-reference. Otherwise it's just way too much data coming back.

fresnizky:
I'll update my PR to a new endpoint

Issue #118, opened 2015-11-20
WvW Objective Upgrade Status


thetoothpick:
Hello,

Would it be possible to get the upgrade status (tier 0/1/2/3) of objectives in WvW (as part of /v2/wvw/matches)?  The upgrades are based on timers now, so they can be _loosely_ calculated based on time captured, but they're also affected by incoming dolyaks, which reduce the timer.  So it'd be useful to get the definite status from the API.

It would also be helpful to get the estimated time of the next upgrade (which would be updated as dolyaks come in), if it's possible.

Something like this:

```
"objectives":[
    {
        "id": "1099-99",
        "type": "Camp",
        "owner": "Red",
        "last_flipped": "2015-11-20T02:51:21Z",
        "claimed_by": "07605FAA-7629-482E-B4AC-4AE673EC0ECE",
        "claimed_at": "2015-11-20T04:19:01Z",
        "upgrades": 0, // or "None"
        "next_upgrade_at": "2015-11-20T03:21:21Z"
    },
    {
        "id": "1099-100",
        "type": "Camp",
        "owner": "Blue",
        "last_flipped": "2015-11-20T04:23:04Z",
        "claimed_by": null,
        "claimed_at": null,
        "upgrades": 3, // or "Tier3" or something
        "next_upgrade_at": null
    },
    ...
]
```

Thanks!

lye:
Yeah, can definitely look into it.

I know the WvW API server has access to the current upgrades, but I'm not sure what the status of upgrade progress is -- if that's not already passed around it might not be very feasible to expose. I also want to expose which guild upgrades are slotted (if any) in each objective, but that's going to take awhile to get out.

thetoothpick:
Any update on this?  The objective upgrades have been changed to be dolyak-based instead of time-based, so it'd be helpful to get the number of dolyaks delivered / remaining too.

lye:
It's still on the backlog; I need a more detailed way of prioritizing tickets I guess.

WvW is probably the next area I'm going to be working on after I finish up with the remainder of the PvP endpoint stuff (and some minor guild endpoint changes). Not sure how long they changes will take to implement, but it's fairly high up on the order of things to do, so there's that at least :<

Rwfrk:
It'd be nice if this were re-prioritized based on the new scoring based on objective tier.

lye:
At long last I think all the bits off this have been implemented.

PR #397, opened 2016-12-03
/v2/wvw/matches: add more objective fields.

### LINKED ISSUE #181: Add guild upgrade details to WvW endpoints.
Basically the current slotted things, tactivation timers, etc.

### LINKED ISSUE #394: Expose points per tick on WvW objectives and matches
Ever since the [WvW scoring changes on November 21st](https://forum-en.guildwars2.com/forum/info/updates/Game-Update-Notes-November-21-2016), it's impossible to predict the amount of points per tick of an objective, map or world.

For reference, the patch notes mention the following PPT:

Objective | Base Points | Secured | Reinforced | Fortified
--- | --- | --- | --- | ---
Camp | 2 | 3 | 4 | 5
Castle | 12 | 18 | 24 | 30
Keep | 8 | 12 | 16 | 20
Tower | 4 | 6 | 8 | 10

This means that it's at least a must to expose the objective tier (#118) so people can calculate the PPT from there. It would be better when every objective in `/v2/wvw/matches` exposes its PPT in case the PPT per tier per objective changes in the future. It would be amazing if the total PPT of the maps and a match can be exposed in `/v2/wvw/matches` as well so we don't even have to do calculations ourselves.

lye:
* points_capture/points_tick since points are variable now.
 * upgrades containing a list of slotted upgrades.

I'm not altogether happy with the output format, very open to suggestions.

Fixes #181; ~doesn't yet expose yaks needed for next tier~.
Fixes #394.
Refs #118. Localized upgrade stuff (and the yaks/tier) needs to be added somewhere.

### DIFF
--- a/v2/wvw/matches.json
+++ b/v2/wvw/matches.json
@@ -64,10 +64,10 @@
 					"owner"          : "Red",
 					"last_flipped"   : "2015-09-18T18:12:29Z",
 					"claimed_by"     : "5AE7196D-F62F-E511-B0A0-0862664D7672",
-					"claimed_at"     : "2015-09-18T18:12:44Z"
+					"claimed_at"     : "2015-09-18T18:12:44Z",
 					"points_tick"    : 2,
 					"points_capture" : 2,
-					"upgrades"       : [123, 134],
+					"guild_upgrades" : [123, 134],
 					"yaks_delivered" : 24
 				},
 				{
@@ -77,7 +77,7 @@
 					"claimed_by"   : null,
 					"claimed_at"   : null,
 					"points_tick"    : 2,
-					"points_capture" : 2,
+					"points_capture" : 2
 				},
 				...
 			]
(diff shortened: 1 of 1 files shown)


thetoothpick:
It'd be nice if you could expose the objective upgrade level (i.e. tier) in the "upgrades" key instead of guild upgrades, and move those to "guild_upgrades".  You can calculate the level based on the number of yaks delivered, but those numbers aren't published in the API anywhere from what I can tell, and hardcoding them would be awkward.

Besides, guild upgrades are a separate concept from objective upgrades, and should be put in a separate key so people don't confuse them.

lye:
I think the yaks/upgrade bits need to go into `/v2/wvw/objectives` since there's some additional localized info for them (icons, names, etc) -- split out into #422.

lye:
Uhh, this went out a bit early and I kind of botched it (didn't get the `upgrades -> guild_upgrades` change).

I should do that real quick.

PR #404, opened 2016-12-09
/v2/pvp/standings -- add skill rating/decay

lye:


lye:
This is [live](https://api.guildwars2.com/v2/pvp/standings?access_token=51969C37-FA95-744C-BB29-075C684EBD33C086D050-124E-4575-994B-304F23DA6F70).

Issue #412, opened 2016-12-19
/v2/wvw/matches endpoint becoming cluttered


mcd1992:
The matches endpoint contains a decent amount of data as is. And will contain more data in the future. See https://github.com/arenanet/api-cdi/pull/397 and https://github.com/arenanet/api-cdi/pull/381

We definitely need to keep the current state of the endpoint as is for backwards compatibility but if possible it might help with the API load to split some of the matches.json endpoint into a few new endpoints in /v2/wvw/

Some scripts have to query the matches.json endpoint multiple times a minute for up to date score/kdr tracking. And in doing so a lot of objective/map data is sent and discarded by the client. I think 4 new endpoints would help with this problem. An endpoint for just scores, global and per map. Endpoint for worlds partaking in this skirmish. Endpoint for kills/deaths global/per-map/per-team. And a endpoint for per map info; could re-use the matches endpoint and just add something like '/v2/wvw/matches/1-1?map=98' or 'matches?world=1001&map=98' ect.

### match-stats
```/v2/wvw/match-stats?id=1-1```
```/v2/wvw/match-stats/1-1```
```/v2/wvw/match-stats?world=1001```

Which will return data similarly to matches.json. Except missing the objectives array and some of the 'static' data that wont change during the skirmish/week. This would also hopefully allow the kills/deaths to update more quickly (instead of at the end of each map tick) like the scores currently do.

### match-scores
```/v2/wvw/match-scores?id=1-1```
```/v2/wvw/match-scores/1-1```
```/v2/wvw/match-scores?id=1-1&map=98```
```/v2/wvw/match-scores?world=1001&map=98```

Which will return score (not kills/deaths/yaks/ect) data pertaining to either all worlds in the match id or just the specified world. Allow 'map=' to filter down to just a specific map in the skirmish.

### match-kd
```/v2/wvw/match-kd?id=1-1```
```/v2/wvw/match-kd/1-1```
```/v2/wvw/match-kd/1-1?map=98```
```/v2/wvw/match-kd?world=1001&map=98```

Which will return kills and deaths (not score/yaks/ect) data pertaining to either all worlds in the match id or just the specified world. Allow 'map=' to filter down to just a specific map in the skirmish.

##
I figure some of this might not be implementable with the current API setup but I'd like to hear what can be implemented. Especially hoping for a endpoint with more quickly updated kills/deaths.

lye:
[...]

PR #416, opened 2016-12-21
Add sub-endpoints for /v2/wvw/matches.

lye:
Three endpoints:

 * /v2/wvw/matches/overview provides mostly static match data.
 * /v2/wvw/matches/scores contains the total and per-map scores.
 * /v2/wvw/matches/stats has the total and per-map KDR bits.

Objective data is the bulk of the payload and can be fetched with the main
endpoint, /v2/wvw/matches.

Addresses most of the points in #412, but doesn't provide `?maps=`
parameter; these response sizes didn't seem large enough to warrant it.

### DIFF
--- a/v2/wvw/matches-overview.js
+++ b/v2/wvw/matches-overview.js
@@ -1,0 +1,32 @@
+// Bulk-expanded sub-endpoint which returns the mostly-static overview
+// data from /v2/wvw/matches.
+
+// GET /v2/wvw/matches/overview
+
+[ "1-1", "1-2" ]
+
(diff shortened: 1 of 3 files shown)


mcd1992:
Looks good, it doesn't say so in the comments but I presume these endpoints will take a worlds parameter? If so then I'm all in. The scores & stats endpoints are my main need since I feel bad querying matches.json and discarding a lot of it.

Thanks!

lye:
Oh, right. Due to an implementation artifact, they totally support the `?world` parameter. I'll add that into the PR so I don't forget in the future.

PR #432, opened 2017-01-11
Require a region to be specified for leaderboards.

lye:
This is a breaking change (leaderboards previously didn't require a region
field -- which returned one region or the other depending on which
datacenter your request was routed to). The old behavior doesn't make much
sense but was maintained just in case someone was depending on it. This
removes that backwards compat.

Planning on deploying this in two weeks, on January 25, 2017.


lye:
> Planning on deploying this in two weeks 

yeah okay

PR #433, opened 2017-01-12
Add the current map id to the mumble link data.

lye:
I meant to add `floor_id` but I got confused.

Well, whatever.


apoch:
As far as I can tell this ID is already available in the link data. Closing just for cleanliness.

PR #435, opened 2017-01-17
Add .skills to /v2/professions.

lye:
Rather than enumerating `.training`, this should pick up all the skills that can be equipped that aren't weapon skills. Also mungling this up is a new `.flags` field which has some boring static data.

This first pass doesn't include things like specialization requirements; will add that in a separate feature.

### DIFF
--- a/v2/professions.js
+++ b/v2/professions.js
@@ -108,8 +108,38 @@
 			"specialization" : 12,
 			"skills" : [ /* ... */ ]
 		}
-	}
+	},
+	// List of all non-weapon skills that can be equipped by this
+	// profession (doesn't include transform skills and such).
+	"skills" : [
+		{
+			"id"   : 34,
+			"slot" : "Profession_1",
+			"type" : "Profession"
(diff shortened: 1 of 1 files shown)


v2/professions.js
```diff
@@ -108,9 +108,39 @@
...
 		}
-	}
+	},
+	// List of all non-weapon skills that can be equipped by this
+	// profession (doesn't include transform skills and such).
+	"skills" : [
+		{
+			"id"   : 34,
+			"slot" : "Profession_1",
+			"type" : "Profession"
```
windwarrior: maybe add `"requires_spec":  "Chronomancer"`?

darthmaim: I would add that as `"specialization": 40`, like it is currently done with weapons.

lye: > This first pass doesn't include things like specialization requirements; will add that in a separate feature.

Need to add in some backend support before I can add those in.

PR #509, opened 2017-04-28
Add WvW guild kdr endpoints.

lye:
Not really happy with the URL structure, open to suggestions.

Also still debating whether to make the guild query require authentication. Right now I'm leaning torwards "no".

### DIFF
--- a/v2/wvw/guild-kdrs.js
+++ b/v2/wvw/guild-kdrs.js
@@ -1,0 +1,76 @@
+// There's a handful of endpoints which serve per-guild KDR statistics.
+// The kills/deaths only count the final killing blow (not kill credits)
+// and only if either (or both) the killer/killee are repping a guild
+// at the time of murder.
+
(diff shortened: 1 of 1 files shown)


rikkuness:
Do you need all 3 of `["red", "green", "blue"]` ? Theoretically one of these values will always be 0 right?

lye:
Many of the guilds actually have values for all three teams -- meaning they've got members repping (and killing) on two or more sides.

rikkuness:
I'd never actually considered that possibility, cool :)

lye:
This is technically live but looks like it's horribly broken. Will open a second issue for it.

rikkuness:
I take it the `top` endpoint and stuff is "live" but the guild specific one still is not?

lye:
Whoops, forgot to turn that on. Should be enabled in ~5 minutes.

PR #514, opened 2017-05-10
Add /v2/commerce/delivery.

### LINKED ISSUE #321: Money & items up for collection in the tradingpost
It would be neat if we got the information of how much money and which items are waiting for the user to collect in the tradingpost. I remember vaguely that there were some issues with that, but its good to track this nevertheless. :)

### LINKED ISSUE #513: Expose gold/items waiting for pickup in the TP.
[Forum ref](https://forum-en.guildwars2.com/forum/community/api/Get-Gold-Items-to-be-Picked-up-at-TP).

lye:
Pretty much what it says on the tin.

Fixes #321 
Fixes #513

### DIFF
--- a/v2/commerce/delivery.js
+++ b/v2/commerce/delivery.js
@@ -1,0 +1,28 @@
+// Authenticated endpoint for exposing item/coins available for
+// pickup at the trading post. Requires "account" and "tradingpost"
+// scopes.
+
+// NOTE: There may be some small race conditions around purchases
+// made before/after you log out which can cause either duplicate 
+// entries or omitted entries, but they should clear up within
+// five minutes. Cache coherency is a hard problem.
+
+// GET /v2/commerce/delivery?access_token=<account+tradingpost>
+
+{
+  "coins": 4294967295,
+  "items": [
+    {
+      "id": 19700,
+      "count": 3
+    },
+    {
+      "id": 19700,
+      "count": 4
+    },
+    {
+      "id": 25557,
+      "count": 1
+    }
+  ]
+}


risingashes:
Is this intended to display items beyond the current 100x UI limit?

No reference to being paginated is included unlike in /v2/commerce/transactions/transactions

for users who consistently have well more than 100x items in their trading post queue this could lead to the same issues that caused transactions to be paginated.

lye:
The current implementation will dump everything.

That uhh. Do you have an account or character name that has a lot of entries? On second thought I may need to consider paginating the results so that my poor machines don't catch fire.

risingashes:
Ashesrising.7840 would likely be my worst one. Full of Barbed Thorns and sadness.

I assume API keys aren't needed by you, but on the off chance I've made a temp one with full perms: 48D46A88-31F3-0D47-9291-2C7C11A82FAA58F462F7-2CC2-4D55-9E33-1BE5F646A8B8

lye:
Awesome, thanks, I'll see if I can pull your account data and see what the damage is. I'm thinking it might be worth while for the API to just combine stacks -- so instead of 99 stacks of 250 thorns it just shows up as one stack of 24750 thorns.

risingashes:
@lye Did you end up rejecting the combination of stacks prior to sending?

Seems that my accounts aren't being handled correctly by any sites incorporating /delivery/ due to the mass amounts of data now being sent.

Is doing so still in the works?

If not, is it possible to give /delivery/ a separate permission so edge cases like myself can prevent the issue?

lye:
I definitely dropped the ball on that; let me see if I can combine them after-the-fact.

